Issue #8, opened 2015-01-06
failed to run godep get to generate go dependency


zhang-hua:
When running ```godep get ./...```, it complains:
```
package github.com/cloudfoundry/bosh-micro-cli/deployer/registry

	imports github.com/cloudfoundry/bosh-micro-cli/deployer/registry

	imports github.com/cloudfoundry/bosh-micro-cli/deployer/registry: cannot find package "github.com/cloudfoundry/bosh-micro-cli/deployer/registry" in any of:

	/usr/local/go/src/pkg/github.com/cloudfoundry/bosh-micro-cli/deployer/registry (from $GOROOT)
	/Users/dongdong/Work/Projects/bosh-agent/src/github.com/cloudfoundry/bosh-micro-cli/deployer/registry (from $GOPATH)
godep: exit status 1 
```

Based on the investigation, we found the only dependency exists in integration test for fake-registry here:
https://github.com/cloudfoundry/bosh-agent/blob/587fefc980298ca993fd70e2024ca7d6b9e45629/integration/fake-registry/fake-registry.go#L7

The actual referenced code does exists here:
https://github.com/cloudfoundry/bosh-agent/tree/master/Godeps/_workspace/src/github.com/cloudfoundry/bosh-micro-cli/deployer/registry
but this is not synchronized with bosh-micro-cli repo which broke the godep.

-IBM pair Edward + Tom

bosh-ci-push-pull:
The registry code was moved within the bosh-micro-cli. The bosh-agent
imports have not yet been updated to reflect that change. You won't be able
to bump those vendored dependencies until that's done.

If you do a godep restore instead, that will check out the vendored
revisions into your go workspace, rather than checking out the latests
revisions.

Or you can fix the import yourself and submit a PR (or do you guys have
commit access?).

-Karl Isenberg
Pivotal BOSH Team

zhang-hua:
ok. we don't have commit access. we will submit a PR for you to review. Thanks!

-IBM pair Edward + Tom

cppforlife:
Closing since PR was merged. Reopen if it's still a problem.

PR #9, opened 2015-01-07
Update the import in fake-registry

xingzhou:
Bump the import to the latest 'registry' code in bosh-micro-cli project.

According to issue #8(https://github.com/cloudfoundry/bosh-agent/issues/8),
the registry code has been updated in bosh-micro-cli project, need to update
related imports.

Signed-off-by: Tom Xing <xingzhou@cn.ibm.com>


xingzhou:
we replaced the old registry code with the lastest code under Godeps/_workspace.

mariash:
Hi @xingzhou, why you needed to update the registry? It is not clear from PR.

xingzhou:
Hi @mariash, when we try to establish the bosh-agent env using `godep get ./...`, we found out that the code has some compilation errors(screenshot is in issue #8), which implies the import in the fake_registry code was using some old version of registry code, so we just submit this PR to bump the vendored dependency to the latest version.

IBM pair Tom & Edward

PR #14, opened 2015-02-13
Multi-homed hosts handle default gateway properly

mmb:
Don't blindly set HasDefaultGateway to true for all networks

[#87982930]

Signed-off-by: Brian Cunnie <cunnie@pivotal.io>


cunnie:
Thanks @mmb — `bosh-agent` had changed so much, including partially fixing this problem, that we finished fixing (i.e. we fixed CentOS's improper configuration of the default gateway).

mmb:
Thanks for fixing this @cunnie

PR #17, opened 2015-04-27
Fix unit tests from ./bin/test-unit script

allomov:
Hey, all. 

Here is a PR to fix unit tests. I guess that the problem was extracting bosh sigar into a separate package. This changes make all unit tests from `./bin/test-unit` script to be green.

Best wishes, 
Alex L.


allomov:
I need to close this PR because I added stuff for testing to `Godeps` folder in this commit. I will open another clean PR soon.

@cppforlife do you think that files from test suites can be added to `.gitignore` to prevent such mistakes? Here is a list of files I added accidentally: https://gist.github.com/allomov/cf8de78fa6d43b77872c.

PR #18, opened 2015-04-27
fix unit tests from ./bin/test-unit script

allomov:
Hey, all.

This PR is copy of https://github.com/cloudfoundry/bosh-agent/pull/17 without test stuff in Godeps folder.

> This is a PR to fix unit tests. I guess that the problem was extracting bosh sigar into a separate package. This changes make all unit tests from ./bin/test-unit script to be green.

Thank you, 
Alex L.


cppforlife:
I believe this has been fixed.

Issue #15, opened 2015-04-24
Can't use File source without registry.


allomov:
You can set the agent to read settings from file using following config options:
```yaml
{
  ...
  "Infrastructure": {
    "Settings": {
      "Sources": [
        {
          "Type": "File",
          "MetaDataPath": "/home/vcap/metadata",
          "UserDataPath": "/home/vcap/userdata",
          "SettingsPath": "/home/vcap/settings"
        }
      ]
    }
  }
}
```
Without `UseRegistry` option you'll get [this error](https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/settings_source_factory.go#L156): `File source is not supported without registry`.

Could you tell if there is any actual reason to restrict to read settings from file if you don't use registry?

bosh-ci-push-pull:
Not technical reason, we just have not implemented that in bosh-agent:
https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/settings_source_factory.go#L156

Our plan is to actually remove registry completely so at some point we will
implement this option.

allomov:
May I create PR for this?

mariash:
Hi @allomov, 

If you want to inject the file with settings you can use the file based registry. The settings source in agent.json represents the metadata source, which in case of file is file based metadata/userdata source. This is to be consistent with other source types.

```
{
  ...
  "Infrastructure": {
    "Settings": {
      "Sources": [
        {
          "Type": "File",
          "MetaDataPath": "/home/vcap/metadata",
          "UserDataPath": "/home/vcap/userdata"
        }
      ]
    }
  }
}
```
In your userdata file just set registry endpoint to the file path.
```
{
  "server": {
    "name": "instance-id"
  },
  "registry": {
    "endpoint": "/home/vcap/settings.json"
  }
}
```

allomov:
Thank you, @mariash, for answer. It may help.

But don't you think that having `SettingsPath` [option](https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/settings_source_factory.go#L53) in file source settings can cause a confusion? Could you add more details about consistency with other source types?

Because from my view, it is easier to use `SettingsPath` option to read settings from file, rather than passing registry endpoint through userdata to read the settings from the same file. Another point for using `SettingsPath` in file source without registry is that registry is going to be deprecated. I guess this changes should be made sometime. What do you think?

cppforlife:
[...]

PR #19, opened 2015-04-27
Add possibility to use File Settings Source without registry.

allomov:
The main goal of this PR is discussed in #15. 

PR contains changes from #18 to fix unit tests.

### DIFF
--- a/infrastructure/file_settings_source.go
+++ b/infrastructure/file_settings_source.go
@@ -10,8 +10,6 @@ import (
 )
 
 type FileSettingsSource struct {
-	metaDataFilePath string
-	userDataFilePath string
 	settingsFilePath string
 
 	fs boshsys.FileSystem
@@ -21,15 +19,11 @@ type FileSettingsSource struct {
 }
 
 func NewFileSettingsSource(
-	metaDataFilePath string,
-	userDataFilePath string,
 	settingsFilePath string,
 	fs boshsys.FileSystem,
 	logger boshlog.Logger,
 ) *FileSettingsSource {
 	return &FileSettingsSource{
-		metaDataFilePath: metaDataFilePath,
-		userDataFilePath: userDataFilePath,
 		settingsFilePath: settingsFilePath,
 
 		fs: fs,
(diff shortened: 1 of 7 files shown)


infrastructure/file_settings_source_test.go
```diff
@@ -0,0 +1,82 @@
...
+	Describe("PublicSSHKeyForUsername", func() {
+		It("returns an empty string", func() {
+			publicKey, err := source.PublicSSHKeyForUsername("fake-username")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(publicKey).To(Equal(""))
+		})
+	})
+
+	Describe("Settings", func() {
+		Context("when the settings file exists", func() {
```
maximilien: What if the file exists but could not be unmarshall, e.g., had bad JSON content? Maybe a test for this specific case could be added?

infrastructure/file_settings_source.go
```diff
@@ -0,0 +1,62 @@
...
+
+	fs boshsys.FileSystem
+
+	logger boshlog.Logger
+	logTag string
+}
+
+func NewFileSettingsSource(
+	metaDataFilePath string,
+	userDataFilePath string,
```
cppforlife: you can remove `metaDataFilePath` and `userDataFilePath` from this class since it only uses settingsFilePath.

allomov: done.

infrastructure/file_settings_source_test.go
allomov: Thank you for your note. I used [file metadata service test](https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/file_metadata_service_test.go#L32-L50) as an example for this test. I added test for invalid format both to file_settings_source_test.go and file_metadata_service_test.go.

maximilien: Nice. Thanks @allomov

dpb587-pivotal:
Closing - this PR has been here for a long time now with no recent activity. If this functionality is still required, please follow up on this PR for more discussion on the continued use case with respect to registry, and we can definitely reopen for discussion.

PR #20, opened 2015-05-18
Warn that HttpsHandler Send() method do nothing

uzzz:
Hi

We had a mistake in our deployment manifest: instead of nats:// schema there was https://, but we wasted a lot of time debugging and reading source until we figured out what actually was the problem. I believe it's useful to warn user if he's trying to send using https handler, that basically do nothing. Probably it's even better to panic, as in RegisterAdditionalFunc function, but I'm not sure about it.

### DIFF
--- a/micro/https_handler.go
+++ b/micro/https_handler.go
@@ -16,6 +16,8 @@ import (
 	boshsys "github.com/cloudfoundry/bosh-agent/system"
 )
 
+const httpsHandlerLogTag = "Https Handler"
+
 type HTTPSHandler struct {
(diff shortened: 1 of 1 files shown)


cppforlife:
This will be overly noisy since agent keeps on using Send() every minute. I wonder if we should output it only once?

jaresty:
merge story [#152974764](https://www.pivotaltracker.com/story/show/152974764)

tjvman:
Please make pull requests for the bosh-agent against the `develop` branch. Rebasing.

PR #22, opened 2015-07-20
Eliminate “root partition must be the first one”

hashmap:
Function findRootDevicePath in platform/linux_platform.go requires
that root partition must be the first one. It can't be satisfied on
some platforms, for example on POWER (ppc64le).

This fix eliminates this requirement by detecting next available
partition id for /swap and /data instead hardcoded 2 and 3
respectively.

### DIFF
--- a/bin/test-unit
+++ b/bin/test-unit
@@ -37,7 +37,7 @@ $bin/test-unhandled-errors
 let "result+=$?"
 
 echo -e "\n Testing packages..."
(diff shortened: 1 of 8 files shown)


hashmap:
I'm a member of cfibmers (IBM) org

hashmap:
I deleted my repo by mistake, sending a new PR, rebased to current develop

PR #23, opened 2015-07-22
Support specifying a list of unmanaged network interfaces

tacgomes:
These patches add support for specifying unmanaged NICs (network interfaces that should be ignored by the Agent) and improve some unit tests.

### DIFF
--- a/agent/bootstrap_test.go
+++ b/agent/bootstrap_test.go
@@ -360,7 +360,7 @@ func init() {
 				arping := bosharp.NewArping(runner, fs, logger, boshplatform.ArpIterations, boshplatform.ArpIterationDelay, boshplatform.ArpInterfaceCheckDelay)
 				interfaceConfigurationCreator := boshnet.NewInterfaceConfigurationCreator(logger)
 
-				ubuntuNetManager := boshnet.NewUbuntuNetManager(fs, runner, ipResolver, interfaceConfigurationCreator, arping, logger)
+				ubuntuNetManager := boshnet.NewUbuntuNetManager(fs, runner, ipResolver, interfaceConfigurationCreator, arping, []string{}, logger)
 
 				ubuntuCertManager := boshcert.NewUbuntuCertManager(fs, runner, logger)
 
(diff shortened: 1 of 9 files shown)


cppforlife:
Interesting. Out of curiosity in your environment who is managing these interfaces?

tacgomes:
Nothing :) On my dual port NIC scenario, one of the ports is connect to a LAN segment that is controlled by OpenStack's Neutron service. As Neutron provides both private and public IP addresses to the machine, I've not use for the second port. Hence, it is not even plugged to any switch.

cppforlife:
Is this something that would be better configured through the CPI or the Director so that you don't have to modify the stemcell?

tacgomes:
I believe that it can't be done on the CPI or director.  I need this functionality during the bootstrap phase of the bosh-agent, whereas it still hasn't received the configuration from the director or the metadata service (for OpenStack clouds).

This is because the first thing being done on the bootstrap phase is to get the public key

https://github.com/cloudfoundry/bosh-agent/blob/master/agent/bootstrap.go#L46

But for retrieving the public key, a minimal network must be set

https://github.com/cloudfoundry/bosh-To receive the configuration from the director, agent/blob/master/infrastructure/http_metadata_service.go#L43

Note that I could make ensureMinimalNetwork to not do any networking setup on bootstrap by adding e.g. "iface eth0 inet dhcp" to /etc/network/interfaces for Ubuntu Openstack images, and then read the configuration to disable some interfaces from the director/CPI, but that would still require to modify the stemcell and in a more hacky way IMO.

cppforlife:
@tacgomes I'm a bit confused ensureMinimalNetwork actually results in "iface eth0 inet dhcp". Why would you have to modify it?

tacgomes:
Hi @cppforlife . See this line of code

https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/http_metadata_service.go#L176

As there aren't configured interfaces (at least on the Ubuntu stemcell), the ensureMinimalNetwork() will call SetupNetwork(), which in turn will run code that will try to DHCP from every physical network interface.

If I added `iface eth0 inet dhcp` to the `/etc/network/interfaces` file, the call to SetupNetworking() would be skipped.

aashah:
This PR has aged quite significantly, I'm gonna close this out. Please feel free to comment if there's more to add!

PR #25, opened 2015-08-13
Support for Softlayer multi-path feature

gu-bin:
This is needed for Bluemix on SoftLayer (SL) since the best disks on SL are iSCSI and the changes are needed to correctly setup these disks. Most of this code was part of a forked copy of the agent code that the Bluemix team has been maintaining so it's already in production but we want to contribute this back.


maximilien:
Hi, all,

@gu-bin is part of the IBM organization and working with me here in Beijing. I will have her added to the `cf-ibmers` shortly. Best,

dr.max

gu-bin:
new PR #29 was created to replace this one.

PR #29, opened 2015-09-10
Support for Softlayer multi-path feature

gu-bin:
This is needed for Bluemix on SoftLayer (SL) since the best disks on SL are iSCSI and the changes are needed to correctly setup these disks. Most of this code was part of a forked copy of the agent code that the Bluemix team has been maintaining so it's already in production but we want to contribute this back.

This new PR changed to use device path resolver infrastructure. The new multipath_device_path_resolver.go is added.


cppforlife:
Is this PR out of date? I see a lot of commits but a tiny diff... may be rebase is needed?

I thought this PR was about adding iscsi attachment logic to the agent so that agent can properly connect to any iscsi device.

maximilien:
@gu-bin Your tracker story is still pointing to this PR. Can you update @cppforlife and myself to the new PR by commenting here as well. Thanks,
max

gu-bin:
@cppforlife I've combined all the commits to one. Please let me know if any other problems. Thanks.

infrastructure/devicepathresolver/multipath_device_path_resolver.go
```diff
@@ -0,0 +1,21 @@
...
+
+func NewMultipathDevicePathResolver(usePreformattedPersistentDisk bool) DevicePathResolver {
+	return multipathDevicePathResolver{usePreformattedPersistentDisk: usePreformattedPersistentDisk}
+}
+
+func (r multipathDevicePathResolver) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
+	if r.usePreformattedPersistentDisk {
+		return diskSettings.Path, false, nil
+	} else {
+		return diskSettings.Path + "-part", false, nil
```
cppforlife: If this is the only change to the Agent at this point, you could just make a change to softlayer CPI to return disk path including "-part". Then we would not have to have this PR.

gu-bin:
@cppforlife @bosh-ci-push-pull
Softlayer can't partition disk for us before use so the parameter "UsePreformattedPersistentDisk" in agent.json must be set to "false" and we need bosh agent to partition and format the disk for us.

When bosh agent do the partition for the disk, it must use the real path of the disk like
/dev/mapper/3600a098038303051555d4637396e4c7a
And after partition, the partition path is something like
/dev/mapper/3600a098038303051555d4637396e4c7a-part1

So the disk path that CPI passes to bosh agent must be something like /dev/mapper/3600a098038303051555d4637396e4c7a with no "-part" appended.
If CPI passes something like "/dev/mapper/3600a098038303051555d4637396e4c7a-part" to bosh agent which is not a real disk path, the bosh agent can't partition the disk. So we still need this PR.

gu-bin:
@cppforlife @bosh-ci-push-pull
Any comments for this PR?

cppforlife:
I am looking through the softlayer cpi code to understand a bit more how this could be consolidated...

gu-bin:
After another review of the code, we found the way to add a new ```devicePathResolver``` to support Softlayer multi-path feature doesn't work for ```MountPersistentDisk``` in ```linux_platform.go```. If use ```multipathDevicePathResolver``` to ```GetRealDevicePath``` that adds "-part" to the device path, the ```Partition``` will fail, because "/dev/mapper/xxxxxx-part" is not a valid disk path to partition. 

The root cause is that ```partitionPath``` is hard coded to ```realPath + "1"```. But for multi-path, the ```partitionPath``` should be ```realPath + "-part1"```. The problem exists in ```MountPersistentDisk```, ```UnmountPersistentDisk``` and ```IsPersistentDiskMounted``` of ```linux_platform.go```. I wonder if we can have something like ```partitionPathResolver``` which can resolve partition path accordingly. Especially for multi-path, it will add "-part1" not "1" for partition path.

gu-bin:
@cppforlife @bosh-ci-push-pull
What's  your comment for this PR?

cppforlife:
@gu-bin Emailed you and Max to get an environment to play around with some of these features.

gu-bin:
@cppforlife @bosh-ci-push-pull
I've sent you the environment info via email.

gu-bin:
@cppforlife @bosh-ci-push-pull
This PR is finalized. Please review. Thanks.

bosh-ci-push-pull:
time.sleep is fine. its location what doesnt make sense to me. lets move it after err check. no amount of time waited will make that err switch its value.

gu-bin:
@cppforlife @bosh-ci-push-pull
I've moved the time.Sleep after err check. Please review.

cppforlife:
Can you also please add analogous unit tests for -part1 just like we have for 1 suffix?

gu-bin:
@cppforlife @bosh-ci-push-pull
Unit tests have been added. My local run of test-unit is successful. Please review.

gu-bin:
@cppforlife @bosh-ci-push-pull
We changed to use retryable to check xxx-part1 partition to show up instead of time.Sleep. Please review. Thanks.

gu-bin:
@cppforlife @bosh-ci-push-pull 
Any update for this PR?

maximilien:
@cppforlife can we get a status on this one, I believe @gu-bin addressed all pending issues.

Thanks,

Max

cppforlife:
oops thx for the bump. will take a look at this one tomorrow.

rnandi:
Merged to develop.

Thanks for your pull request!

PR #32, opened 2015-09-17
Change user (vcap) and root password separately

mattcui:
If the root password is not defined in yml, it will be not updated, same as the way to set vcap password.


platform/fakes/fake_platform.go
```diff
@@ -197,8 +199,14 @@ func (p *FakePlatform) SetupSSH(publicKey, username string) error {
 	return p.SetupSSHErr
 }
 
-func (p *FakePlatform) SetUserPassword(user, encryptedPwd string) (err error) {
-	p.UserPasswords[user] = encryptedPwd
+func (p *FakePlatform) SetUserPassword(user string, encryptedPwd string) (err error) {
+	if user == boshsettings.VCAPUsername {
+		p.UserPassword = encryptedPwd
+	}
```
maximilien: minor: new line between `if` statements

mattcui: I added a new line there, as discussed with @maximilien and @cfdreddbot , we need to escalate the issue about disabling root login to high level operation or management team.

maximilien: @mattcui, I am in Raleigh all of next week and will see face-to-face all the key Bluemix folks. I plan to raise this issue with them as well as others. Let's sync at scrum on this before I leave on Sunday.

mattcui: @maximilien sure, in the status update call with Fabio last night, I mentioned you will stop by him to talk something about OS reload and others...:smile: ,  maybe plus different size issue for stemcell, they want to create an additional 100G stemcell, I am thinking if we could set stemcell size in deployment yaml like openstack....

maximilien: @mattcui yes, I'll add story for adding stemcell size

mattcui:
@cfdreddbot @maximilien  For long-term, we should consider to disable root, but we still need this change before it happens.

settings/settings.go
```diff
@@ -99,12 +99,17 @@ type Env struct {
...
+	return e.Bosh.UserPassword
+}
+
+func (e Env) GetRootPassword() string {
+	return e.Bosh.RootPassword
 }
 
 type BoshEnv struct {
-	Password string `json:"password"`
+	UserPassword string `json:"user_password"`
```
cppforlife: cannot rename `password` to `user_password` since that's not backwards compatible.

agent/bootstrap.go
```diff
@@ -116,19 +120,28 @@ func (boot bootstrap) Run() (err error) {
...
 	}
 
-	err = boot.platform.SetUserPassword(boshsettings.VCAPUsername, password)
+	return nil
+}
+
+func (boot bootstrap) setRootPasswords(env boshsettings.Env) error {
+	password := env.GetRootPassword()
+	if password == "" {
+		return nil
```
cppforlife: i dont think we should check for emptiness. if the goal is to stop resetting root password then we should just add a boolean configuration `keep_root_password` on BoshEnv so that root password is not changed.

maximilien:
@mattcui can you update this to address comments?

cppforlife:
Looks good I think. Can you rebase this against bosh-agent. There are some differences that would not exist if rebased.

settings/settings.go
```diff
@@ -240,7 +245,8 @@ func (n Network) IsVIP() bool {
 //	},
 //	"env": {
 //		"bosh": {
-//			"password": null
+//			"userpassword": null,
+//          "rootpassword": null
```
cppforlife: this comment update is not in sync with changes.

agent/bootstrap.go
```diff
@@ -121,14 +125,30 @@ func (boot bootstrap) setUserPasswords(env boshsettings.Env) error {
...
 	err := boot.platform.SetUserPassword(boshsettings.RootUsername, password)
 	if err != nil {
-		return bosherr.WrapError(err, "Setting root password")
+		return bosherr.WrapError(err, "Setting vcap password")
+	}
+
+	return nil
+}
+
+func (boot bootstrap) setRootPasswords(env boshsettings.Env) error {
```
cppforlife: After reading this code it doesnt look like it properly works for vcap users anymore. I think something like this would be easier to understand:

```
func (boot bootstrap) setUserPasswords(env boshsettings.Env) error {
	password := env.GetPassword()
	if password == "" {
		return nil
	}

	err := boot.platform.SetUserPassword(boshsettings.VCAPUsername, password)
	if err != nil {
		return bosherr.WrapError(err, "Setting vcap password")
	}

	if !env.KeepRootPassword() {
		err := boot.platform.SetUserPassword(boshsettings.RootUsername, password)
		if err != nil {
			return bosherr.WrapError(err, "Setting root password")
		}
	}

	return nil
}
```

and remove SetRootPasswords

mattcui: Yes, it's still working in progress. I will post a comment to ask you for review after I complete all code change and unit test. Thanks.

mattcui:
@cppforlife The code change is ready for review and merge,  please let me know if there is any problem. Thanks.

mattcui:
@cppforlife I believe it's already rebased, Please review. Thanks.

PR #35, opened 2015-09-25
add network attribute "preconfigured" to support Softlayer network

gu-bin:
This is for code review and no unit test yet. If the code is acceptable, I will add unit test code. Thanks.


cppforlife:
This is tricky... Ubuntu typically handles the resolv.conf changes, since it reserves the right to change it. I dont think if preconfigured flag is set, agent should modify any networking properties. If softlayer is messing around with VM internal files it should then also configure DNS.

gu-bin:
@cppforlife @bosh-ci-push-pull 
We are using PowerDNS to provide DNS resolution between the VMs in a deployment. So we need to add the bosh director's IP to each VM's nameserver list. Softlayer only configures its own default DNS which can't resolve the special domain names provided by the PowerDNS.

gu-bin:
@cppforlife @bosh-ci-push-pull 
Any more comments?

cppforlife:
Did you guys find out if it's possible to use DHCP on softlayer?

gu-bin:
I opened a ticket to Softlayer asking if they support DHCP and the answer is no. We need to setup our own DHCP server if we need to use DHCP and I think it's not possible for now.

Here is the answer from Softlayer:
The configuration you have suggested is not supported. Our network uses VLAN's to separate our larger network into smaller spaces. Our network does allow this capability to set up your own DHCP server and allow other devices which can use that DHCP server to get their IP address. Can you please elaborate further on why you need this? If you set this up then all configuration of this would need to be done from your end. In most cases this is not necessary and causes a lot of unnecessary configuration and another point of failure.

gu-bin:
@cppforlife @bosh-ci-push-pull
Any more comments?

cppforlife:
Fair enough no DHCP. 

Separate question: if you write resolv.conf manually, what happens when network is restarted for some reason? Does that file change, because from our prev experiences that's a standard ubuntu behavior.

gu-bin:
Dmitriy,

When network is restarted, resolv.conf does not change. It keeps what bosh 
agent writes into it.


Best Regards,

Bin Gu (顾彬)

Bluemix, IBM China Software Development Lab
Tel: (8610)82454944
Email: gubin@cn.ibm.com
Address: 3F, Ring Bldg, ZhongGuanCun Software Park No.28, Dongbeiwang West 
Road No.8,ShangDi, Haidian District,Beijing 100193, PRC

voelzmo:
That is interesting. From what the [manpages say](http://manpages.ubuntu.com/manpages/precise/en/man8/resolvconf.8.html), `ifdown` followed by `ifup` should rewrite `resolv.conf` according to the information present in `/etc/network/interfaces`

gu-bin:
@cppforlife @bosh-ci-push-pull
Any more comments?

cppforlife:
All right. After thinking sometime I think we can go forward. I've adapted your code as a prototype into something similar for warden cpi: https://github.com/cppforlife/bosh-agent/commit/4ffbfa23ab99c6b49de31c29ba5c9a8e0fe71320. Except in warden cpi case I did not implement resolv.conf change. If you take this commit and add resolv.conf changes via resolvconf I that will satisfy your case.

Regarding how to add info to resolv.conf, we should do the following:
- write out selected DNS servers to /etc/resolvconf/resolv.conf.d/head
- run `resolvconf`(http://manpages.ubuntu.com/manpages/lucid/man8/resolvconf.8.html) which should generate resolv.conf file based on director configuration.

gu-bin:
@cppforlife @bosh-ci-push-pull
I've added resolv.conf changes via resolvconf as you suggested and created a new PR to the cppforlife bosh-agent repository. Here is the new PR.
https://github.com/cppforlife/bosh-agent/pull/1

Close this one for now.

cppforlife:
Looks good. Keep in mind that you should rebase necessary changes (all preconfigured changes) against this repo instead of my own fork.

gu-bin:
@cppforlife @bosh-ci-push-pull
I've committed all the preconfigured changes to this repo and also finished the unit test code. Please review. Thanks.

gu-bin:
@cppforlife @bosh-ci-push-pull
Any more comments for this PR?

cppforlife:
@gu-bin nope. looks good. we will pull this in.

mariash:
merged to develop

cppforlife:
In 3138 build.

Issue #16, opened 2015-04-26
Why do you return empty string as a name in file metadata source?


allomov:
[In this code](https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/file_metadata_service.go#L67) empty string is return as a server name. I thought that file metadata source works much a like as another metadata services. So if you set `Name` option, then agent can get it from any type of source. Think that this should be changed.

maximilien:
If I recall correctly in the `FileMetaData` the `ServerName` does not make sense but we need to implement so that the interface method requirement is satisfied. A better question could be why keep that method in that interface?

allomov:
I guess the reason is that BOSH writes VM name as a metadata (link to [sources](https://github.com/cloudfoundry/bosh/blob/master/bosh_cli_plugin_micro/lib/bosh/deployer/instance_manager.rb#L124)). 

In case with OpenStack CPI bosh-agent reads VM name from metadata and then asks registry about initial settings using the name as an ID (link to [sources](https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/http_registry.go#L42)).

allomov:
https://github.com/cloudfoundry/bosh-agent/pull/39 should fix this problem

PR #39, opened 2015-10-19
Add Azure support.

AbelHu:


allomov:
Some time ago I came across the same issue. I was surprised that `GetServerName` in `FileMetadataSource` returned empty string https://github.com/cloudfoundry/bosh-agent/issues/16. It's nice that you fixed it.

infrastructure/file_metadata_service.go
```diff
@@ -46,6 +45,10 @@ func (ms fileMetadataService) GetPublicKey() (string, error) {
 }
 
 func (ms fileMetadataService) GetInstanceID() (string, error) {
+	if ms.metaDataFilePath == "" {
+		return "", nil
+	}
+
```
allomov: I am sure that this condition is redundant. If file path is not set, or set to the wrong path, `ReadFile ` is going to return the error ((source)[https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/file_metadata_service.go#L51]) and bosh-agent will fail. Failing in this case is expected behaviour.

platform/disk/sfdisk_partitioner.go
```diff
@@ -50,6 +50,7 @@ func (p sfdiskPartitioner) Partition(devicePath string, partitions []Partition)
 
 	_, _, _, err := p.cmdRunner.RunCommandWithInput(sfdiskInput, "sfdisk", "-uM", devicePath)
 	if err != nil {
+		p.cmdRunner.RunCommandWithInput("d\n2\nd\nw\n", "fdisk", devicePath)
```
allomov: Is this code related to the problem this PR solves?

AbelHu: This is used to fix an issue we hit on Azure. I will separate them to two PRs. Thanks.

infrastructure/file_metadata_service.go
AbelHu: If  the failing is expected behavior. I can remove it.

allomov: `GetInstanceID` returns error if something is wrong, the idea is that someone who calls this function is responsible for processing this case. In the same time, missing `metaDataFilePath` is an error and it will be processed by `ReadFile`, so I guess the condition should be removed.

platform/disk/sfdisk_partitioner.go
AbelHu: @allomov I separate them into two commits but they are still included in one PR. Is it OK?

allomov: @AbelHu I think it would be better if you'll create another PR to add this line. Lets wait for comments from @cppforlife or @maximilien.

mariash:
PR was merged here (without disk reformatting for now): https://github.com/cloudfoundry/bosh-agent/commit/7d0e933b71736c56d238c5a9969ee969b1258337

PR #42, opened 2015-10-29
Eliminate “root partition must be the first one”

hashmap:
Function findRootDevicePath in platform/linux_platform.go requires
that root partition must be the first one. It can't be satisfied on
some platforms, for example on POWER (ppc64le).

This fix eliminates this requirement by detecting next available
partition id for /swap and /data instead hardcoded 2 and 3
respectively.

Signed-off-by: Alexander Lomov <alexander.lomov@altoros.com>


allomov:
@hashmap I believe you need to change my account to @allomov, it has signed CLA. is it hard to do it?

PR #44, opened 2015-11-09
User can cancel while drain agent action is running

hashmap:
This fix makes drain action cancellable. It also introduces
CancellableScript interface and its implementation for drain and
parallel scripts.

Small change in fake_cmd_runner was made to propely test parallel async
execution of scripts.

[#106642152] (https://www.pivotaltracker.com/story/show/106642152)

Signed-off-by: Yulia Gaponenko <yulia.gaponenko@ru.ibm.com>

### DIFF
--- a/agent/action/concrete_factory_test.go
+++ b/agent/action/concrete_factory_test.go
@@ -37,6 +37,7 @@ var _ = Describe("concreteFactory", func() {
 		jobScriptProvider boshscript.JobScriptProvider
 		factory           Factory
 		logger            boshlog.Logger
+		cancelCh          chan struct{}
 	)
 
 	BeforeEach(func() {
@@ -51,6 +52,7 @@ var _ = Describe("concreteFactory", func() {
 		specService = fakeas.NewFakeV1Service()
 		jobScriptProvider = &fakescript.FakeJobScriptProvider{}
 		logger = boshlog.NewLogger(boshlog.LevelNone)
+		cancelCh = make(chan struct{}, 1)
 
 		factory = NewFactory(
 			settingsService,
@@ -82,7 +84,8 @@ var _ = Describe("concreteFactory", func() {
 	It("drain", func() {
 		action, err := factory.Create("drain")
 		Expect(err).ToNot(HaveOccurred())
-		Expect(action).To(Equal(NewDrain(notifier, specService, jobScriptProvider, jobSupervisor, logger)))
+		// Cannot do equality check since channel is used in initializer
+		Expect(action).To(BeAssignableToTypeOf(DrainAction{}))
 	})
 
 	It("fetch_logs", func() {
--- a/agent/action/concrete_factory_test.go
+++ b/agent/action/concrete_factory_test.go
@@ -37,7 +37,6 @@ var _ = Describe("concreteFactory", func() {
 		jobScriptProvider boshscript.JobScriptProvider
 		factory           Factory
 		logger            boshlog.Logger
-		cancelCh          chan struct{}
 	)
 
 	BeforeEach(func() {
@@ -52,7 +51,6 @@ var _ = Describe("concreteFactory", func() {
 		specService = fakeas.NewFakeV1Service()
 		jobScriptProvider = &fakescript.FakeJobScriptProvider{}
 		logger = boshlog.NewLogger(boshlog.LevelNone)
-		cancelCh = make(chan struct{}, 1)
 
 		factory = NewFactory(
 			settingsService,
--- a/agent/action/drain.go
+++ b/agent/action/drain.go
@@ -85,10 +85,10 @@ func (a DrainAction) Run(drainType DrainType, newSpecs ...boshas.V1ApplySpec) (i
 
 	script := a.jobScriptProvider.NewParallelScript("drain", scripts)
 
-	resultsChan := make(chan error)
-	go func() { resultsChan <- script.Run() }()
+	resultsCh := make(chan error, 1)
+	go func() { resultsCh <- script.Run() }()
 	select {
-	case result := <-resultsChan:
+	case result := <-resultsCh:
 		a.logger.Debug(a.logTag, "Got a result")
 		return 0, result
 	case <-a.cancelCh:
--- a/agent/action/drain.go
+++ b/agent/action/drain.go
@@ -18,8 +18,9 @@ type DrainAction struct {
 	specService       boshas.V1Service
 	jobSupervisor     boshjobsuper.JobSupervisor
 
-	logTag string
-	logger boshlog.Logger
+	logTag   string
+	logger   boshlog.Logger
+	cancelCh chan struct{}
 }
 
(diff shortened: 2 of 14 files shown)


agent/action/concrete_factory_test.go
```diff
@@ -51,6 +52,7 @@ var _ = Describe("concreteFactory", func() {
 		specService = fakeas.NewFakeV1Service()
 		jobScriptProvider = &fakescript.FakeJobScriptProvider{}
 		logger = boshlog.NewLogger(boshlog.LevelNone)
+		cancelCh = make(chan struct{}, 1)
```
cppforlife: doesnt seem that this channel is used.

agent/action/concrete_factory_test.go
```diff
@@ -37,6 +37,7 @@ var _ = Describe("concreteFactory", func() {
 		jobScriptProvider boshscript.JobScriptProvider
 		factory           Factory
 		logger            boshlog.Logger
+		cancelCh          chan struct{}
```
cppforlife: doesnt seem that this channel is used.

agent/action/drain.go
```diff
@@ -81,9 +83,18 @@ func (a DrainAction) Run(drainType DrainType, newSpecs ...boshas.V1ApplySpec) (i
 		scripts = append(scripts, script)
 	}
 
-	parallelScript := a.jobScriptProvider.NewParallelScript("drain", scripts)
-
-	return 0, parallelScript.Run()
+	script := a.jobScriptProvider.NewParallelScript("drain", scripts)
+
+	resultsChan := make(chan error)
```
cppforlife: let's make this guy buffered of size 1 so that goroutine can be released if cancelling is done before script finishes running. 

stylistic tip: use 'Ch' suffix for channels.

agent/action/drain_test.go
```diff
@@ -327,4 +327,43 @@ var _ = Describe("DrainAction", func() {
...
+					"foo": boshas.PackageSpec{
+						Name: "foo",
+						Sha1: "foo-sha1-new",
+					},
+				},
+			}
+		)
+
+		BeforeEach(func() {
+
```
cppforlife: whitespace

agent/action/drain_test.go
```diff
@@ -327,4 +327,43 @@ var _ = Describe("DrainAction", func() {
...
+			parallelScript = &fakescript.FakeCancellableScript{}
+			jobScriptProvider.NewDrainScriptStub = func(jobName string, params boshdrain.ScriptParams) boshscript.CancellableScript {
+				return fakedrain.NewFakeScript("fake-tag")
+			}
+			jobScriptProvider.NewParallelScriptReturns(parallelScript)
+			currentSpec := boshas.V1ApplySpec{}
+			specService.Spec = currentSpec
+
```
cppforlife: whitespace

agent/action/drain_test.go
```diff
@@ -327,4 +327,43 @@ var _ = Describe("DrainAction", func() {
...
+		})
+
+		Context("when action was not canceled yet", func() {
+			It("cancel action", func() {
+
```
cppforlife: whitespace

agent/action/drain_test.go
```diff
@@ -327,4 +327,43 @@ var _ = Describe("DrainAction", func() {
...
+				_, err := action.Run(DrainTypeShutdown, newSpec)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = action.Cancel()
+				Expect(err).ToNot(HaveOccurred())
+			})
+		})
+
```
cppforlife: whitespace

agent/script/drain/concrete_script.go
```diff
@@ -96,12 +108,30 @@ func (s ConcreteScript) runOnce(params ScriptParams) (int, error) {
...
+	var result boshsys.Result
+
+	// Can only wait once on a process but cancelling can happen multiple times
+	for processExitedCh := process.Wait(); processExitedCh != nil; {
+		select {
+		case result = <-processExitedCh:
+			processExitedCh = nil
+		case <-s.cancelCh:
+			// Ignore possible TerminateNicely error since we cannot return it
+			process.TerminateNicely(10 * time.Second)
```
cppforlife: let's log the error similar to https://github.com/hashmap/bosh-agent/blob/cancel-drain-action/agent/action/run_errand.go#L89-L91

agent/script/drain/concrete_script.go
```diff
@@ -96,12 +108,30 @@ func (s ConcreteScript) runOnce(params ScriptParams) (int, error) {
...
+			return 0, bosherr.Error("Script was cancelled by user request")
```
cppforlife: i dont think we need to return here since process will be terminated and it will return some exit code.

agent/script/parallel_script.go
```diff
@@ -66,6 +66,23 @@ func (s ParallelScript) Run() error {
 	return s.summarizeErrs(passedScripts, failedScripts)
 }
 
+func (s ParallelScript) Cancel() error {
+	s.logger.Debug(s.logTag, "Canceling a parallel script")
+	existingScripts := s.findExistingScripts(s.allScripts)
+	for _, script := range existingScripts {
+		if script, ok := script.(CancellableScript); ok {
```
cppforlife: should we just add Cancel to the Script interface. For non drain scripts we can implement it as always returning an error for now?

agent/script/parallel_script.go
```diff
@@ -66,6 +66,23 @@ func (s ParallelScript) Run() error {
...
+			s.logger.Debug(s.logTag, "Canceling a script")
```
cppforlife: this log doesnt look that useful.

agent/script/parallel_script.go
```diff
@@ -66,6 +66,23 @@ func (s ParallelScript) Run() error {
...
+			err := script.Cancel()
+			if err != nil {
+				return bosherr.Errorf("Failed to cancel %s: %s", s.name, err)
```
cppforlife: may be `bosherr.WrapErrorf(err, "'%s' script did not cancel", s.name)`?

agent/script/drain/concrete_script.go
hashmap: @cppforlife in this case line 136 will return an error, which looks a bit misleading to me,

agent/script/parallel_script.go
hashmap: Perhaps, but I like that we intentionally specify that script is cancellable without cluttering all Script implementations. Would be nice to have default implementation in interface like in Java 8.

benmoss:
Sorry to bump an old PR but this modified bosh-utils' `fake_cmd_runner` directly in bosh-agent instead of in that repo. I just synced bosh-utils with the latest changes and luckily the ConcreteScript tests failed and was able to trace it back. We should be more careful about this in the future.

PR #48, opened 2015-11-25
Photon Adding a new scsi-id device resolver

dingyin:
This PR contains a scsi-id device resolver. It will be used to discover disk using scsi-id.

### DIFF
--- a/infrastructure/devicepathresolver/scsi.go
+++ b/infrastructure/devicepathresolver/scsi.go
@@ -1,0 +1,33 @@
+package devicepathresolver
+
+import (
+	boshsettings "github.com/cloudfoundry/bosh-agent/settings"
+)
+
+type scsi struct {
+	scsiVolumeID 	DevicePathResolver
+	scsiID          DevicePathResolver
+}
+
+func NewScsi(
+	scsiVolumeID DevicePathResolver,
+	scsiID DevicePathResolver,
+) DevicePathResolver {
+	return scsi{
+		scsiVolumeID: scsiVolumeID,
+		scsiID: scsiID,
+	}
+}
+
+func (scsiResolver scsi) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
+	var realPath string
+	var timeout bool
+	var err error
+
+	if len(diskSettings.ID) > 0 {
+		realPath, timeout, err = scsiResolver.scsiID.GetRealDevicePath(diskSettings)
+	} else if len(diskSettings.VolumeID) > 0 {
+		realPath, timeout, err = scsiResolver.scsiVolumeID.GetRealDevicePath(diskSettings)
+	}
+	return realPath, timeout, err
+}
--- a/infrastructure/devicepathresolver/scsi.go
+++ b/infrastructure/devicepathresolver/scsi.go
@@ -2,6 +2,7 @@ package devicepathresolver
 
 import (
 	boshsettings "github.com/cloudfoundry/bosh-agent/settings"
+	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 )
 
 type scsi struct {
@@ -28,6 +29,9 @@ func (scsiResolver scsi) GetRealDevicePath(diskSettings boshsettings.DiskSetting
 		realPath, timeout, err = scsiResolver.scsiID.GetRealDevicePath(diskSettings)
 	} else if len(diskSettings.VolumeID) > 0 {
 		realPath, timeout, err = scsiResolver.scsiVolumeID.GetRealDevicePath(diskSettings)
+	} else {
+		return "", timeout, bosherr.WrapError(err, "Neither ID nor VolumeID exists.")
 	}
+
 	return realPath, timeout, err
 }
--- a/infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
+++ b/infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
@@ -36,7 +36,7 @@ func NewScsiIDDevicePathResolver(
 
 func (idpr scsiIDDevicePathResolver) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
 	if diskSettings.ID == "" {
-		return "", false, bosherr.WrapError("Disk ID is not set")
+		return "", false, bosherr.Errorf("Disk ID is not set")
 	}
 
 	hostPaths, err := idpr.fs.Glob("/sys/class/scsi_host/host*/scan")
@@ -61,7 +61,7 @@ func (idpr scsiIDDevicePathResolver) GetRealDevicePath(diskSettings boshsettings
 		idpr.logger.Debug(idpr.logTag, "Waiting for device to appear")
 
 		if time.Now().After(stopAfter) {
-			return "", true, bosherr.WrapError("Timed out getting real device path for '%s'", diskSettings.ID)
+			return "", true, bosherr.Errorf("Timed out getting real device path for '%s'", diskSettings.ID)
 		}
 
 		time.Sleep(100 * time.Millisecond)
--- a/infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
+++ b/infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
@@ -12,7 +12,7 @@ import (
 
 // Resolves device path by performing a SCSI rescan then looking under
 // "/dev/disk/by-id/*uuid" where "uuid" is the cloud ID of the disk
-type scsiIDDevicePathResolver struct {
+type SCSIIDDevicePathResolver struct {
 	diskWaitTimeout time.Duration
 	fs              boshsys.FileSystem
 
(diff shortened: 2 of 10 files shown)


platform/provider.go
```diff
@@ -90,6 +90,8 @@ func NewProvider(logger boshlog.Logger, dirProvider boshdirs.Provider, statsColl
 		devicePathResolver = devicepathresolver.NewVirtioDevicePathResolver(idDevicePathResolver, mappedDevicePathResolver, logger)
 	case "scsi":
 		devicePathResolver = devicepathresolver.NewScsiDevicePathResolver(500*time.Millisecond, fs)
+	case "scsiid":
+		devicePathResolver = devicepathresolver.NewScsiIDDevicePathResolver(50000*time.Millisecond, fs, logger)
```
cppforlife: Did we find a reason why it was taking so long?

infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+		logger:          logger,
+	}
+}
+
+const logTag = "scsiIDresolver"
+
+func (idpr scsiIDDevicePathResolver) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
+	if diskSettings.ID == "" {
+		err := bosherr.Errorf("Disk ID is not set")
+		idpr.logger.Error(logTag, err.Error())
```
cppforlife: agent doesnt log errors returned since they will be logged at the action handler level. you can also then inline err into the return statement.

infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+		return "", false, err
+	}
+
+	pattern := `^\w{8}(-\w{4}){3}-\w{12}$`
+	match, _ := regexp.MatchString(pattern, diskSettings.ID)
```
cppforlife: i'm not sure if we actually care about the format of the id passed in.

infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+	logger boshlog.Logger,
+) DevicePathResolver {
+	return scsiIDDevicePathResolver{
+		diskWaitTimeout: diskWaitTimeout,
+		fs:              fs,
+		logger:          logger,
+	}
+}
+
+const logTag = "scsiIDresolver"
```
cppforlife: let's put this tag into the struct e.g. https://github.com/cloudfoundry/bosh-agent/blob/master/agent/script/parallel_script.go#L28

infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+		if time.Now().After(stopAfter) {
+			err := bosherr.Errorf("Timed out getting real device path for '%s'", diskSettings.ID)
+			idpr.logger.Error(logTag, err.Error())
+			return "", true, err
+		}
+
+		time.Sleep(100 * time.Millisecond)
+
+		uuid := strings.Replace(diskSettings.ID, "-", "", -1)
+		disks, err := idpr.fs.Glob("/dev/disk/by-id/*" + uuid)
```
cppforlife: should this be `'/dev/disk/by-id/*-' + uuid` to separate off uuid from subsystem?

platform/provider.go
dingyin: There are some activities of hypervisor and guest OS. I am pretty sure it is related to slow hardware or busy system because I can't repo it in our test environment. We are gathering more logs to understanding what activities take this long.

infrastructure/devicepathresolver/scsi_id_device_path_resolver.go
dingyin: We try to strip off the trailing '-' from the ID to match what is under /dev/disk/by-id

cppforlife: Right but isn't the format for by-id/ folder is something like `by-id/ata-blahblbhabblah`?

platform/provider.go
cppforlife: Interesting, is this the only change for the agent? What else is different between this and vsphere stemcell?

dingyin: sphere use scsi path resolver. Here there are something happened to update the scsi-id to match disk-id seeing from hypervisor

cppforlife:
Updates look good. 

I was thinking we should probably combine two scsi resolvers in this way so that we can just reuse vsphere stemcells.

- rename scsi_device_path_resolver to scsi_volume_id_device_path_resolver
- add new scsi_device_path_resolver that does something like this:
  - if diskSettings.ID is non-empty use scsi_id_device_path_resolver
  - if diskSettings.VolumeID is non-empty use scsi_volume_id_device_path_resolver
  - if both ID and VolumeID are non-empty return an error
  - if none are provided return an error

Sounds good?

settings/settings.go
```diff
@@ -72,10 +72,12 @@ func (s Settings) PersistentDiskSettings(diskID string) (DiskSettings, bool) {
 			if hashSettings, ok := settings.(map[string]interface{}); ok {
 				diskSettings.Path = hashSettings["path"].(string)
 				diskSettings.VolumeID = hashSettings["volume_id"].(string)
+				diskSettings.ID = hashSettings["id"].(string)
```
cppforlife: test is missing for checking what happens when `id` is found or not found.

settings/settings.go
```diff
@@ -72,10 +72,12 @@ func (s Settings) PersistentDiskSettings(diskID string) (DiskSettings, bool) {
...
 			} else {
 				// Old CPIs return disk path (string) or volume id (string) as disk settings
 				diskSettings.Path = settings.(string)
 				diskSettings.VolumeID = settings.(string)
+				diskSettings.ID = settings.(string)
```
cppforlife: i dont think we should do this here.

settings/settings.go
```diff
@@ -89,6 +91,7 @@ func (s Settings) EphemeralDiskSettings() DiskSettings {
 	return DiskSettings{
 		VolumeID: s.Disks.Ephemeral,
 		Path:     s.Disks.Ephemeral,
+		ID:       s.Disks.Ephemeral,
```
cppforlife: same thing here, if s.Disks.Ephemeral is a map[string]interface{} let's try to extract id, otherwise let's not set it just like for persistent disks.

settings/settings.go
```diff
@@ -86,10 +91,20 @@ func (s Settings) PersistentDiskSettings(diskID string) (DiskSettings, bool) {
...
 func (s Settings) EphemeralDiskSettings() DiskSettings {
-	return DiskSettings{
-		VolumeID: s.Disks.Ephemeral,
-		Path:     s.Disks.Ephemeral,
+	diskSettings := DiskSettings{}
+
+	if hashSettings, ok := s.Disks.Ephemeral.(map[string]interface{}); ok {
+		diskSettings.Path = hashSettings["path"].(string)
+		diskSettings.VolumeID = hashSettings["volume_id"].(string)
+		diskSettings.ID = ""
```
cppforlife: shouldnt this be hashSettings["id"].(string)?

settings/settings.go
```diff
@@ -72,10 +75,12 @@ func (s Settings) PersistentDiskSettings(diskID string) (DiskSettings, bool) {
 			if hashSettings, ok := settings.(map[string]interface{}); ok {
 				diskSettings.Path = hashSettings["path"].(string)
 				diskSettings.VolumeID = hashSettings["volume_id"].(string)
+				diskSettings.ID = diskID
```
cppforlife: shouldnt this be diskSettings.ID = hashSettings["id"].(string)?

molteanu:
Any update on the status of this PR?

cppforlife:
waiting for @dingyin to update with changes as we discussed.

infrastructure/devicepathresolver/scsi.go
```diff
@@ -0,0 +1,33 @@
...
+	var realPath string
+	var timeout bool
+	var err error
+
+	if len(diskSettings.ID) > 0 {
+		realPath, timeout, err = scsiResolver.scsiID.GetRealDevicePath(diskSettings)
+	} else if len(diskSettings.VolumeID) > 0 {
+		realPath, timeout, err = scsiResolver.scsiVolumeID.GetRealDevicePath(diskSettings)
+	}
+	return realPath, timeout, err
```
cppforlife: let's return an error if none of the ifs match.

cppforlife:
Added a comment about error. Otherwise looks good.

infrastructure/devicepathresolver/scsi.go
```diff
@@ -0,0 +1,37 @@
...
+		scsiID: scsiID,
+	}
+}
+
+func (scsiResolver scsi) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
+	var realPath string
+	var timeout bool
+	var err error
+
+	if len(diskSettings.ID) > 0 {
```
mariash: Since we always setting ID for persistent disk in disk settings, that means we always going to resolve path by ID.

settings/settings.go
mariash: We use diskSettings.ID to store disk CID on line 73. It is better to not share the same property for disk ID and disk CID.

mariash:
@molteanu we will take care of comments above.

PR #49, opened 2015-11-30
Use ntpd instead of ntpdate

xingzhou:
[#107791384]

Signed-off-by: Zhang Hua zhuadl@cn.ibm.com


jaresty:
merge story [#152974725](https://www.pivotaltracker.com/story/show/152974725)

dpb587-pivotal:
We are currently working on a new stemcell line for Ubuntu Xenial which changes the underlying NTP/NTPD implementation, so I'm going to close this in favor of that and since we're not planning on changing the Ubuntu Trusty behavior at this point.

zachgersh:
just to add more detail here - we switched to chrony (comparable with ntpd)

voelzmo:
Can chrony run in a mode which doesn't require additional ports? This is imho the main reason to use ntpdate instead of ntpd for the trusty stemcells.

zachgersh:
Don’t really understand - doesn’t ntpdate have to use a port to stay
synchronized to its NTP server? Are you talking about not persistently
having a port open? Also - this Chrony change was again only for Xenial
unless we were talking about doing a backport but I haven’t heard any of
that discussed.

voelzmo:
@zachgersh ntpd runs in daemon mode and requires to always have a port open on that machine. On the bosh stemcells we therefore decided against ntpd, and instead call the ntpdate binary in regular intervals. I'd be fine if chrony runs in a daemon mode similar to ntpd, I just want to avoid having "yet another port"(tm) open on a stemcell if possible.

PR #51, opened 2015-12-11
Retry to format disk when an error occurs.

AbelHu:


AbelHu:
Close and create a PR to the correct branch develop.

PR #52, opened 2015-12-14
Retry to format disk when an error occurs.

AbelHu:


AbelHu:
Since the old one has been merged, this should be closed.

PR #53, opened 2015-12-16
Get unit tests (almost) all passing on Windows

benmoss:
This fixes the project so that the unit tests almost all pass when run on a Windows machine. The main issue, as you can see from the changes made, is that the `filepath` package will insert OS-specific path separators, ie `/` on unix-likes and `\` on Windows. We will ideally be able to move all of the OS specific filepath manipulation into the OsFileSystem.

Right now the tests pass because we use the FakeFileSystem in almost all our tests. The only exception is the app_test, which uses the OsFileSystem and also starts up a real go-sigar goroutine which panics currently on Windows. That test has just been marked as pending on Windows.

I was thinking we could merge this now rather than keeping it on the Windows branch just to avoid having to resolve conflicts from it, and since it shouldn't effect behavior on other platforms.

I wasn't able to get the integration tests running, so do need to verify that they pass.


benmoss:
Closing this one since I realized it's going to master instead of develop

PR #56, opened 2015-12-23
Linux formatter: return error when failed to detect partition type

### LINKED ISSUE #26: Agent should fail when it does not recognize the type of partition vs formatting it
The bosh-agent should be as defensive as possible, especially when it performs destructive actions. In the current code in `linux_formatter.go` there is a method to check the type of a partition, pasted here:

``` go
func (f linuxFormatter) partitionHasGivenType(partitionPath string, fsType FileSystemType) bool {
    stdout, _, _, err := f.runner.RunCommand("blkid", "-p", partitionPath)
    if err != nil {
        return false
    }

    return strings.Contains(stdout, fmt.Sprintf(` TYPE="%s"`, fsType))
}
```

If the `blkid` call fails (for whatever reasons) instead of returning false which in the caller will result in agent potentially partitioning/formatting the disk which will destroy all data.

Instead the agent should be defensive and propagate failure and thus avoid destroying data in disk.

uzzz:
This PR fixes #26 

Does it look good?

### DIFF
--- a/platform/disk/linux_formatter.go
+++ b/platform/disk/linux_formatter.go
@@ -21,7 +21,12 @@ func NewLinuxFormatter(runner boshsys.CmdRunner, fs boshsys.FileSystem) Formatte
 }
 
 func (f linuxFormatter) Format(partitionPath string, fsType FileSystemType) (err error) {
-	if f.partitionHasGivenType(partitionPath, fsType) {
+	hasGivenType, err := f.partitionHasGivenType(partitionPath, fsType)
+	if err != nil {
+		return bosherr.WrapError(err, "Checking partition type")
+	}
+
+	if hasGivenType {
 		return
 	}
 
@@ -45,11 +50,11 @@ func (f linuxFormatter) Format(partitionPath string, fsType FileSystemType) (err
 	return
 }
 
-func (f linuxFormatter) partitionHasGivenType(partitionPath string, fsType FileSystemType) bool {
+func (f linuxFormatter) partitionHasGivenType(partitionPath string, fsType FileSystemType) (bool, error) {
 	stdout, _, _, err := f.runner.RunCommand("blkid", "-p", partitionPath)
 	if err != nil {
-		return false
+		return false, bosherr.WrapError(err, "Shelling out to blkid")
 	}
 
-	return strings.Contains(stdout, fmt.Sprintf(` TYPE="%s"`, fsType))
+	return strings.Contains(stdout, fmt.Sprintf(` TYPE="%s"`, fsType)), nil
 }
--- a/platform/disk/linux_formatter_test.go
+++ b/platform/disk/linux_formatter_test.go
@@ -1,6 +1,8 @@
 package disk_test
 
 import (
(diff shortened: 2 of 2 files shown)


dpb587-pivotal:
Hi @uzzz - thanks for the PR! We were testing the changes here and discovered what may be an issue. When looking at a disk/partition which has not been formatted (e.g. the first time a device is attached), the `blkid` call will exit with an exit code of 2. This causes `Format` to immediately return the wrapped error and it won't ever end up reaching the `mkswap`/`mke2fs`-formatting system calls (i.e. it won't format blank disks).

We were wondering if you were able to test this patched agent in your own environment and if this patch was working for you. If so, perhaps we need to update our testing environment or look into how this is behaving differently between your environment and ours.

We also noticed is that if we use a disk/partition which has been formatted with a non-ext2/non-swap (e.g. NTFS, filesystem type 0x7), `blkid` will still run and exit successfully. This causes `partitionHasGivenType` to return `false, nil`, but `Format` will still end up reformatting the disk and potentially losing data. I think this was one of the main scenarios #26 was trying to avoid. Do you have any additional insight into how this PR fixes that scenario?

Thanks!

@dpb587-pivotal, @cunnie

dpb587-pivotal:
Hi @uzzz,

We'd be grateful if you could enhance your pull request to do the following:

* use `sfdisk -l /dev/disk-device-path` to make sure there's not an existing partion table on the partition to be formatted. We believe this was the situation that triggered the original issue.  We are agnostic about the mechanism of determining the existence of a partition table (don't `sfdisk` if you don't want).
* If `blkid` exits with a 2, it means that the disk is most likely blank and is okay to proceed with formatting
* if `blkid` returns with a non-ext4 type (e.g. ` TYPE=ext3`), then we do *not* want to proceed with formatting—there is already a filesystem on that partition.

Let us know if you have questions—ping us on the PR once you've pushed changes some we can take another look.

Thanks for the work!

@dpb587-pivotal, @cunnie

cppforlife:
cc @maximilien @jianqiu @mattcui

maximilien:
Hey, @jianqiu and @mattcui,

We should enhance @uzzz's PR with Danny's suggestion and resubmit.

That should probably fix the format persistent disk problem.

uzzz:
Wow, I have missed Danny's comments (probably because of New Year's holidays) and completely forgot about this PR, sorry.

I'll update this PR with Danny's suggestions.

dpb587-pivotal:
It looks like we did indeed make some of the improvements in that commit - thanks for looking that up.

The agent still doesn't do much in terms of protecting against valid but unexpected partition types. For anybody interested in tackling that, feel free to make a new PR.

maximilien:
@mattcui and @gu-bin since we have been bit by this issue a few times on SoftLayer I think we should own the proactive enhancement that @dpb587-pivotal is mentioning in comment above.

I would like us to discuss this in a Beijing next week and add any necessary work items to address as part of the v3 effort. Thoughts?

Thanks,

Max

/cc @cppforlife

PR #58, opened 2016-01-08
Add default sender to nats request

aemengo:
When the agent tries to respond to a NATS message without a sender, the
action will fail without any discernible leads as to where the fault
lies. It typically responds with an: 'Unknown Protocol Operation'

This is just a more friendly interaction.

[#111105094]


cppforlife:
That looks like violation of a contact. I think it's better if agent raises an error saying that sender is not specified instead of polluting mbus with messages.

aemengo:
Now, I'm worried about other kinds of messages that the Agent receives. If some tasks are sent to the agent that don't necessarily need a reply (like a 'cleanup' task), I wouldn't want the task to be halted prematurely.

jaresty:
merge story [#152974697](https://www.pivotaltracker.com/story/show/152974697)

PR #65, opened 2016-02-02
Make change to partition in bootstrap and mount disk call.

kaleo211:


agent/action/mount_disk.go
```diff
@@ -2,37 +2,42 @@ package action
...
+	"fmt"
 
+	boshdpresolv "github.com/cloudfoundry/bosh-agent/infrastructure/devicepathresolver"
 	boshsettings "github.com/cloudfoundry/bosh-agent/settings"
 	boshdirs "github.com/cloudfoundry/bosh-agent/settings/directories"
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
+	boshlog "github.com/cloudfoundry/bosh-utils/logger"
 )
 
+const logTag string = "mount_disk_action"
```
cppforlife: let's change this to be "MountDiskAction"

platform/platform_interface.go
```diff
@@ -47,7 +47,9 @@ type Platform interface {
 	MigratePersistentDisk(fromMountPoint, toMountPoint string) (err error)
 	GetEphemeralDiskPath(diskSettings boshsettings.DiskSettings) string
 	IsMountPoint(path string) (result bool, err error)
+	FindDeviceMatchingMountPoint(string) (string, bool, error)
```
cppforlife: talking with @tylerschultz about this interface, let's try to keep platform interface minimal. we should probably have IsMountPoint return (string, bool, error). wdyt? also name is not necessarily related to devices, since other things could be mounted.

PR #66, opened 2016-02-04
Pr persistent partition and mount disk

kaleo211:


agent/action/mount_disk.go
```diff
@@ -59,11 +64,23 @@ func (a MountDiskAction) Run(diskCid string) (interface{}, error) {
...
+	}
+
+	devicePath, isMountPoint, err := a.diskMounter.IsMountPoint(mountPoint)
 	if err != nil {
 		return nil, bosherr.WrapError(err, "Checking mount point")
 	}
+	a.logger.Info(logTag, "realPath = %s, devicePath = %s, isMountPoint = %s", realPath, devicePath, isMountPoint)
+
 	if isMountPoint {
+		if realPath == devicePath || fmt.Sprint(realPath+"1") == devicePath {
```
mariash: @kaleo211 what is the purpose of comparing devicePath with realPath+1? It is unclear from code and there are no tests that are failing when removing this condition.

kaleo211: in our case, we have a fixed persistent disk, and it always be /dev/sdb. And bosh agent always mounts /dev/sdb1 to /var/vcap/store for the first time. when we want to remount, we want to know sdb is already mounted by that line.

mariash: If bosh-agent always mount it as `/dev/sdb1` why there is a first condition?

kaleo211: we keep that for other cpi, I don't know whether it make sense, :))

mariash: Partitioning is related to OS and should not depend on which CPI it is running on. So it is unclear how it should be affected by other cpi...

kaleo211: I see, we will remove that condition and rerun the integration test,

mariash:
Hi @kaleo211,

We are seeing test failure: `platform/dummy_platform_test.go:97`. We are going to fix it as part of merge. Just want to let you know so that you can improve your tests run in future.

Thanks.

Issue #57, opened 2015-12-28
cloudstack compatible ssh-key metadata URI


poblin-orange:
While working on the cloudstack stemcell support for bosh, met this issue.
https://github.com/cloudfoundry-community/bosh-cloudstack-cpi-release/issues/14

The metadata location for public ssh keys is harcoded in bosh-agent, uses http://<metadata-server>/latest/meta-data/public-keys/0/openssh-key
https://github.com/cloudfoundry/bosh-agent/blob/1dca3244702c18bf2c36483c529d4e7b3fb92b2e/infrastructure/http_metadata_service.go

In cloudstack, the URL to use is http://<metadata-server>/latest/meta-data/public-keys

Could we provide a way in agent.json to specify the correct ssh key URI ? Or fallback to trying /latest/meta-data/public-keys

cppforlife:
What exactly is the format for `/latest/meta-data/public-keys`?

poblin-orange:
here the curl to list cloudstack available metadata :

```
root@0d9834ab-9277-41e0-8fe6-8a12611c5c29:/var/vcap/bosh# curl -X GET http://10.234.242.14/latest/meta-data/
service-offering
availability-zone
local-ipv4
local-hostname
public-ipv4
public-hostname
instance-id
vm-id
public-keys
cloud-identifier
```

And this curl gives directly the public key :

```
root@0d9834ab-9277-41e0-8fe6-8a12611c5c29:/var/vcap/bosh# curl -X GET http://10.234.242.14/latest/meta-data/public-keys
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQCprNYHUTGotiHIzeaUSKvA+WENRvPSN7iv/TrGHF2DQa330Wg4HvQCiC8hm0veYwerGo3+WtcL/DOFYF8dRJF/ELbx6dey31HRREYMlypNXWQpru5guyXGpYor7XegDo+PvBQZDAPqNkj6CBKxotFfSU1YZAggUKAkFDaPCtuSAQ==
```

cppforlife:
I think it makes sense for the agent to try other URL if first URL doesnt not work.

poblin-orange:
It seems this PR will fix this issue https://github.com/cloudfoundry/bosh-agent/pull/71

poblin-orange:
closing, as agent.json can now be configured to find cloudstack specific ssh key url

PR #71, opened 2016-03-30
Allow HTTP Metadata URLs and virtio device prefixes to be configurable

ljfranklin:
**Warning:** this should be merged at the same time as https://github.com/cloudfoundry/bosh/pull/1189 to avoid breaking AWS and OpenStack stemcell building.

Since the metadata service in Google Compute Engine has different paths than AWS or OpenStack for the instance user data, we need to allow the agent to take in the instance user data path, the instance ID path, and the SSH keys path in agent env HTTP sources.

Also, GCE uses "google" as the virtio device prefix, so we must allow this to be configurable as well.

### DIFF
--- a/infrastructure/devicepathresolver/id_device_path_resolver.go
+++ b/infrastructure/devicepathresolver/id_device_path_resolver.go
@@ -11,19 +11,24 @@ import (
 	boshsys "github.com/cloudfoundry/bosh-utils/system"
 )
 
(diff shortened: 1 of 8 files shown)


poblin-orange:
Nice feature. I guess this would also fix https://github.com/cloudfoundry/bosh-agent/issues/57

PR #72, opened 2016-04-12
Replace hardcoded device number to value returned by root device finder

kalambet:
Since for `ppc64le` architecture device number is not `1` by default the code use value returned from root device finder

Signed-off-by: Yulia Gaponenko <yulia.gaponenko@ru.ibm.com>


cppforlife:
@barthy1 huh. i thought this was already changed. can you add a simple unit test to lock this behaviour down?

PR #73, opened 2016-04-18
New Settings Source: Instance metadata

evandbrown:
This change creates a settings source named `InstanceMetadata` that retrieves and unmarshals BOSH settings from a URI on the instance's metadata store. Google Compute Engine supports mutable instance metadata and this will be used in conjunction with the [Google CPI](https://github.com/cloudfoundry-incubator/bosh-google-cpi-release/tree/metadata-registry) to eliminate the registry when using Google Cloud Platform.

### DIFF
--- a/app/config_test.go
+++ b/app/config_test.go
@@ -52,6 +52,12 @@ var _ = Describe("LoadConfigFromPath", func() {
 					  {
 					  	"Type": "CDROM",
 					  	"FileName": "/fake-file-name"
+					  },
+					  {
+						"Type": "InstanceMetadata",
+						"URI": "/fake-uri",
+						"Headers": {"fake": "headers"},
+						"SettingsPath": "/fake-settings-path"
 					  }
 				  ],
 				  "UseServerName": true,
@@ -91,6 +97,11 @@ var _ = Describe("LoadConfigFromPath", func() {
 						},
 						boshinf.CDROMSourceOptions{
 							FileName: "/fake-file-name",
+						},
+						boshinf.InstanceMetadataSourceOptions{
+							URI:          "/fake-uri",
+							Headers:      map[string]string{"fake": "headers"},
+							SettingsPath: "/fake-settings-path",
 						},
 					},
 					UseServerName: true,
--- a/app/config_test.go
+++ b/app/config_test.go
@@ -195,6 +195,23 @@ var _ = Describe("LoadConfigFromPath", func() {
 		Expect(err.Error()).To(ContainSubstring("Unmarshalling source type 'HTTP'"))
 	})
 
+	It("returns errors if failed to decode InstanceMetadata source options", func() {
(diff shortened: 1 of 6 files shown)


evandbrown:
When this is resolved, it will need to be enabled in the BOSH stemcell building code, similar to https://github.com/evandbrown/bosh/commit/d604cea531fcc993d5dab4ec9fdcf4e9132daa1c

evandbrown:
Added https://github.com/cloudfoundry/bosh/pull/1226 to make sure this is pulled into BOSH core when ready.

infrastructure/settings_source_factory.go
```diff
@@ -199,6 +218,10 @@ func (s *SourceOptionsSlice) UnmarshalJSON(data []byte) error {
 				var o HTTPSourceOptions
 				err, opts = mapstruc.Decode(m, &o), o
 
+			case optType == "InstanceMetadata":
```
cppforlife: let's update the test in https://github.com/cloudfoundry/bosh-agent/blob/master/app/config_test.go to make sure we dont lose this new configuration.

infrastructure/instance_metadata_settings_source.go
```diff
@@ -0,0 +1,66 @@
...
+	settingsPath    string
+
+	platform boshplatform.Platform
+	logger   boshlog.Logger
+
+	logTag          string
+	metadataService DynamicMetadataService
+}
+
+func NewInstanceMetadataSettingsSource(
```
cppforlife: i think we should add new _test.go for this class based on https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/http_metadata_service_test.go.

infrastructure/settings_source_factory.go
evandbrown: D'oh, totally skipped the tests.

Fixed in c94b00a8ead774c970fa449e89b8f392975ccef5

infrastructure/instance_metadata_settings_source.go
evandbrown: c94b00a8ead774c970fa449e89b8f392975ccef5

app/config_test.go
```diff
@@ -195,6 +195,23 @@ var _ = Describe("LoadConfigFromPath", func() {
...
+					}]
+				}
+			}
+		}`)
+
+		_, err := LoadConfigFromPath(fs, "/fake-config.conf")
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("Unmarshalling source type 'InstanceMetadata'"))
+	})
+
```
ljfranklin: Might also expand the "returns populates config" test at the top of the file to check that "InstanceMetadata" source can be successfully unmarshaled.

evandbrown: Good call. Done.

cppforlife:
Closing, this was merged.

PR #75, opened 2016-04-20
bosh-agent for windows!

benmoss:


cppforlife:
- do we need to always absolute-ize paths for symlink to work? and if so would it be better to hide that in the file system class?

- imho a better abstraction for script commands would be to windows specific implementation into wrapped script runner instead of forcing everyone to use new script command factory? similar to how LinuxBindMounter hides its bind mounting operations (https://github.com/cloudfoundry/bosh-agent/blob/master/platform/disk/linux_bind_mounter.go). this way boshsys.Command remains to be same.

- do we need to make `RunPackagingCommand` public? seems like a internaly detail of the compiler

- replace `Describe("Testing with Ginkgo"` with something more meaningful

- replace "N/A" load values with an empty `[]string{}`

- should `SystemMemStats/SystemPageStats/UsedDiskSpace` methods be moved into gosigar since that's the library that is responsible for different OS stats?

- why do we need `bosh-utils/state` package? and if we do need it can it be localized to the package that uses it.

mhoran:
Thanks for the review, @cppforlife.

> do we need to always absolute-ize paths for symlink to work? and if so would it be better to hide that in the file system class?

Relative symlinks do not work reliably in Windows. We've moved this functionality to the system package.

> imho a better abstraction for script commands would be to windows specific implementation into wrapped script runner instead of forcing everyone to use new script command factory? similar to how LinuxBindMounter hides its bind mounting operations (https://github.com/cloudfoundry/bosh-agent/blob/master/platform/disk/linux_bind_mounter.go). this way boshsys.Command remains to be same.

We've created a [story](https://www.pivotaltracker.com/story/show/118186197) to address this.

> do we need to make RunPackagingCommand public? seems like a internaly detail of the compiler

Privatized!

> replace Describe("Testing with Ginkgo" with something more meaningful

There are a number of vendored packages with tests that have this description, but we did rename one Describe that was modified in one of our commits. If you had a specific one you wanted us to change, please let us know.

> replace "N/A" load values with an empty []string{}

Returning `[]string{}` breaks the bosh CLI for `bosh vms --vitals`, which does not expect a nil load value. It does work for `bosh instances --vitals`.

Removing the `omitempty` from the load key of the vitals struct is gracefully handled by the bosh CLI, but is a change in behavior that we were not comfortable making. It also breaks the healthcheck unit test.

Instead, we use `[]string{""}`, which does not change behavior and does not break the CLI. Please let us know if you'd prefer something else.

> should SystemMemStats/SystemPageStats/UsedDiskSpace methods be moved into gosigar since that's the library that is responsible for different OS stats?

We've [created](https://www.pivotaltracker.com/story/show/118200857) a story to address this.

> why do we need bosh-utils/state package? and if we do need it can it be localized to the package that uses it.

Removed.

xoebus:
Neat. I saw that there are quite a few binary files that have been checked in. Are these necessary?

benmoss:
@xoebus @cppforlife 

>Neat. I saw that there are quite a few binary files that have been checked in. Are these necessary?

We got rid of the tar files, the remaining binaries are executables that we need for tests.

> should SystemMemStats/SystemPageStats/UsedDiskSpace methods be moved into gosigar since that's the library that is responsible for different OS stats?

Moved into `gosigar`.

> imho a better abstraction for script commands would be to windows specific implementation into wrapped script runner instead of forcing everyone to use new script command factory? similar to how LinuxBindMounter hides its bind mounting operations (https://github.com/cloudfoundry/bosh-agent/blob/master/platform/disk/linux_bind_mounter.go). this way boshsys.Command remains to be same.

Do you mind if we defer this work and put it into a chore? It's a bit of a bigger change than the other ones and we'd rather get this PR merged.

cppforlife:
> the remaining binaries are executables that we need for tests.

Is it possible to compile them on the fly? We have concerns that we (or others) dont know where this binaries come and what they contain. How do we rebuild them?

> Do you mind if we defer this work and put it into a chore?

After talking with Tyler, we agreed it's fine. Let's make sure that that chore is high up to do this after PR is merged.

sbenario:
The chore is in the backlog as #118186197. It's currently relatively high
up (before the next release marker)

re: binaries
    compiling on the fly would be really difficult for some of them. Since
these are binaries for tests, is that acceptable? We could perhaps add a
chore to have them built in the CI pipeline if that would be better

tylerschultz:
My concern with the test binaries is that I don't know where they came from, what they are, or how to rebuild them. I like the idea of having the pipeline rebuild them, at least I'd have some idea about how to proceed. 

Would it make sense to put in another chore to build the binaries in a/the pipeline, and once the pipeline is built, remove the binaries from the repo?

sbenario:
Created story #118507493 to track removal of binaries (to either be
compiled on the fly or as part of the CI pipeline).

tylerschultz:
Ok, one last request. :-)

Please rebase and squash. Then champagne. 🎉

PR #79, opened 2016-05-19
Increase disk resolver timeout for Bracker computing

krishnanrs:
At 500ms, the device resolver timeout is too aggressive for some cloud deployments.

PR #80, opened 2016-06-01
Bracket computing

krishnanrs:
Provide a configurable device path resolver timeout

For some clouds, a device path resolver timeout of 500ms is too aggressive.
Providing a configurable way of being able to define it in the agents.json file.

### DIFF
--- a/agent/action/fakes/fake_clock.go
+++ b/agent/action/fakes/fake_clock.go
@@ -20,6 +20,14 @@ type FakeClock struct {
 	sleepArgsForCall []struct {
 		d time.Duration
 	}
+	SinceStub        func(t time.Time) time.Duration
+	sinceMutex       sync.RWMutex
+	sinceArgsForCall []struct {
+		t time.Time
+	}
+	sinceReturns struct {
+		result1 time.Duration
+	}
 	NewTimerStub        func(d time.Duration) clock.Timer
 	newTimerMutex       sync.RWMutex
 	newTimerArgsForCall []struct {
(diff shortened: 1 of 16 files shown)


krishnanrs:
Hello, I haven't received any responses or comments on this yet. Let me know if there are any questions or concerns regarding this change.

cppforlife:
@krishnanrs Hmm. I've been thinking more about this change and looking over a few code paths. I've reopened https://github.com/cloudfoundry/bosh-agent/pull/79 and I think we'll pull that change in instead of making this configurable.

cppforlife:
Will close this one.

krishnanrs:
Dimitri, appreciate you looking into this and agreeing to a change for setting a larger timeout value. However I still think that there is a value in making this timeout configurable.

Is there a possibility to accept both changes? I can re-submit the original change with a lower timeout value (maybe 5 secs. instead of 30 secs.). 500 ms. is definitely too aggressive in my opinion.

cppforlife:
@krishicks one of our goals is eventually to get to a "generic" stemcell which means that bosh-agent should work out of the box with the same configuration for all iaas-es (slowly getting there...), so adding additional agent options would actually be going the other way. 

given that we are going to use higher value for the timeout as you proposed, why do you want to make this value configurable?

krishicks:
@cppforlife I think you meant to tag @krishnanrs.

^_^

krishnanrs:
@cppforlife providing additional agent options avoid overriding hard-coded values, especially given that the go-agent needs to be compiled for all supported OS. Hence my preference is to use a slightly higher default value and also providing agent options.

PR #86, opened 2016-08-10
fixes formatting using parted on centOS

xtreme-nitin-ravindran:
Signed-off-by: Tanzeeb Khalili <tkhalili@pivotal.io>

`parted -m` exits with 0 on centos. Fix checks stderr instead of error objects.


dpb587-pivotal:
Merged - thanks!

Issue #83, opened 2016-07-22
bosh agent doesn't clean /var/vcap/data/root_tmp when upgrade from old stemcell


gu-bin:
This problem should only happen in Softlayer as we do os-reload to upgrade stemcell. We open this issue to see if community can help solve it. Thanks.

We are trying to upgrade stemcell from 3169 to 3262.2. During the upgrade, the vm bosh agent failed to start due to 

```
2016-07-21_06:55:19.43133 [main] 2016/07/21 06:55:19 ERROR - App setup Running bootstrap: Setting up tmp dir: Creating root tmp dir: Running command: 'mkdir -p /var/vcap/data/root_tmp', stdout: '', stderr: 'mkdir: cannot create directory '/var/vcap/data/root_tmp': File exists
```

This is because in old stemcell 3169, /var/vcap/data/root_tmp exists in the ephemeral disk like this:

```
-rwx------ 1 root root 134217728 Jul 21 08:38 /var/vcap/data/root_tmp*
```

During stemcell upgrade, we do os-reload in Softlayer and keep using the existing ephemeral disk. So the file /var/vcap/data/root_tmp is still there.

But in new stemcell 3262.2, bosh agent changes to create a directory with the same name at the same place. As the ephemeral disk is not recreated, bosh agent handles it as an existing one and doesn't clean it. As a result, bosh agent failed to create the directory /var/vcap/data/root_tmp and fail to start.

If possible, would you please fix it like this:
Before run `mkdir -p /var/vcap/data/root_tmp`, check if there is a file with the same name and path existing. If yes, remove it. Thanks.

/cc @maximilien @jianqiu @mattcui

maximilien:
Should be a straight forward change to `linux_platform.go` in [lines 700-703](https://github.com/cloudfoundry/bosh-agent/blob/8a1a382bf63892f833935c1cf79afc4bdd3fb78b/platform/linux_platform.go#L700-L703)

from

```
err := p.fs.MkdirAll(boshTmpDir, tmpDirPermissions)
if err != nil {
	return bosherr.WrapError(err, "Creating temp dir")
}
```

to something like

```
_, err := os.Stat(boshTmpDir)
if !os.IsNotExist(err) {
  err = os.RemoveAll(boshTmpDir)
  if err != nil {
	  return bosherr.WrapError(err, "Removing temp dir")
  }
}

err := p.fs.MkdirAll(boshTmpDir, tmpDirPermissions)
if err != nil {
[...]

PR #91, opened 2016-08-29
Bosh agent will scrub ephemeral disk when upgrading stemcell

zhanggbj:
Bosh agent will scrub ephemeral disk when upgrading stemcell

[issue #83](https://github.com/cloudfoundry/bosh-agent/issues/83)

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -526,6 +526,47 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string) error {
 		return bosherr.WrapError(err, "Creating data dir")
 	}
 
+	if p.options.ScrubEphemeralDisk {
+		agentVersionFilePath := path.Join(p.dirProvider.DataDir(), ".bosh", "agent_version")
+		stemcellVersionFilePath := path.Join(p.dirProvider.EtcDir(), "stemcell_version")
+		stemcellVersion, err := p.fs.ReadFileString(stemcellVersionFilePath)
+		if err != nil {
+			return bosherr.WrapError(err, "Reading stemcell version file")
+		}
+
+		if !p.fs.FileExists(agentVersionFilePath) {
+			err = p.fs.WriteFileString(agentVersionFilePath, stemcellVersion)
+			if err != nil {
+				return bosherr.WrapError(err, "Writting agent version file")
+			}
+		} else {
+			agentVersion, err := p.fs.ReadFileString(agentVersionFilePath)
+			if err != nil {
+				return bosherr.WrapError(err, "Reading agent version file")
+			}
+
+			if agentVersion != stemcellVersion {
+				sysRunPath := path.Join(p.dirProvider.DataDir(), "sys", "run")
+				_, err = p.diskManager.GetMounter().Unmount(sysRunPath)
+				if err != nil {
+					return bosherr.WrapErrorf(err, "Unmounting '%s'", sysRunPath)
+				}
+
+				for _, content := range contents {
+					err = p.fs.RemoveAll(content)
+					if err != nil {
+						return bosherr.WrapErrorf(err, "Removing '%s'", content)
+					}
+				}
+
+				err = p.fs.WriteFileString(agentVersionFilePath, stemcellVersion)
+				if err != nil {
+					return bosherr.WrapErrorf(err, "Updating agent version file '%s'", agentVersionFilePath)
+				}
+			}
+		}
+	}
+
 	var swapPartitionPath, dataPartitionPath string
 
 	// Agent can only setup ephemeral data directory either on ephemeral device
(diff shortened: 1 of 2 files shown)


zhanggbj:
Hi @maximilien , @mattcui 

This is ready for review.  Thank you!

platform/linux_platform.go
```diff
@@ -569,6 +573,49 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string) error {
...
+
+			if agentVersion != stemcellVersion {
+				sysRunPath := path.Join(p.dirProvider.DataDir(), "sys", "run")
+				_, err = p.diskManager.GetMounter().Unmount(sysRunPath)
+				if err != nil {
+					return bosherr.WrapErrorf(err, "Unmounting '%s'", sysRunPath)
+				}
+
+				dataGlob := path.Join(p.dirProvider.DataDir(), "*")
+				contents, err := p.fs.Glob(dataGlob)
```
cppforlife: err should be checked.

platform/linux_platform.go
```diff
@@ -569,6 +573,49 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string) error {
 		return bosherr.WrapError(err, "Mounting data partition")
 	}
 
+	if p.options.ScrubEphemeralDisk {
```
cppforlife: i think we should move this up to line #528.

platform/linux_platform.go
```diff
@@ -569,6 +573,49 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string) error {
...
+			}
+		} else {
+			agentVersion, err := p.fs.ReadFileString(agentVersionFilePath)
+			if err != nil {
+				return bosherr.WrapError(err, "Reading agent version file")
+			}
+
+			if agentVersion != stemcellVersion {
+				sysRunPath := path.Join(p.dirProvider.DataDir(), "sys", "run")
+				_, err = p.diskManager.GetMounter().Unmount(sysRunPath)
```
cppforlife: i dont think we need to umount sys/run.

platform/linux_platform.go
```diff
@@ -569,6 +573,49 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string) error {
 		return bosherr.WrapError(err, "Mounting data partition")
 	}
 
+	if p.options.ScrubEphemeralDisk {
+		agentVersionFilePath := path.Join(p.dirProvider.DataDir(), ".bosh", "agent_version")
+		stemcellVersionFilePath := path.Join(p.dirProvider.BaseDir(), "bosh", "etc", "stemcell_version")
```
cppforlife: dirProvider. EtcDir()

platform/linux_platform.go
```diff
@@ -569,6 +573,49 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string) error {
 		return bosherr.WrapError(err, "Mounting data partition")
 	}
 
+	if p.options.ScrubEphemeralDisk {
+		agentVersionFilePath := path.Join(p.dirProvider.DataDir(), ".bosh", "agent_version")
```
cppforlife: dirProvider.BoshDir()

platform/linux_platform.go
zhanggbj: It can't be deleted directly as it is mounted.
```
# df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            3.9G  4.0K  3.9G   1% /dev
tmpfs           802M  208K  802M   1% /run
tmpfs           1.0M     0  1.0M   0% /var/vcap/data/sys/run
```

platform/linux_platform.go
zhanggbj: Done

platform/linux_platform.go
zhanggbj: Just noticed it has been globed in line #510, so removed it.

platform/linux_platform.go
zhanggbj: I think it is better to keep agent_version in a hidden dir `/var/vcap/data/.bosh/agent_version`  as it is only generated and used when ScrubEphemeralDisk is true. What do you think?

zhanggbj:
Hi @cppforlife,

Comments addressed. Would you please help to take a look at the updates? Thank you.

platform/linux_platform.go
cppforlife: SetupDataDir() gets called after SetupEphemeralDisk() so there is no way it should be mounted at this time.

zhanggbj: Yes, right, if we are using a brand new disk, `/var/vcap/data/sys/run`  dir is setup after SetupEphemeralDisk() and this part of code will just write agent version file and do nothing else. I'm thinking when we are upgrading stemcell,  after mounting `/var/vcap/data`,  the directories under it have been setup as we are using the old disk, so still need to umount `/var/vcap/data/sys/run` and then delete it, right? Or am I missing something here? Thank you.

zhanggbj: Hi @cppforlife, I'm still a little bit confusing with the situation mentioned in last comments, when upgrading stemcell and using the old disk, `/var/vcap/data/sys/run` should have been mounted. Any suggestions? Thank you .

cppforlife: when upgrading a stemcell, machine is powered off, nothing will be mounted when first boostrap happens with a new os.

zhanggbj: Thank you @cppforlife for your explanation:)  I removed it and all comments have been addressed I think, please help to take another review when you got time. Thanks a lot.

cppforlife:
Story: https://www.pivotaltracker.com/story/show/129727797

zhanggbj:
Hi @cppforlife ,

I've addressed all of the comments, would you please help to take another review? Thank you:)

PR #96, opened 2016-09-09
Retry getting agent settings

Kiemes:
Make agent code more resilient, e.g. to handle temporary network issues

[#129348495](https://www.pivotaltracker.com/story/show/129348495)

Signed-off-by: Felix Riegger <felix.riegger@sap.com>

PR #102, opened 2016-10-25
Retry agent settings 257.x 3263.x

beyhan:
This change is already applied to master with this [pull request](https://github.com/cloudfoundry/bosh-agent/pull/96). We would like to have it merged here as well while we wait to the next stemcell release. We build an agent version with this change and verified that the retry logic works. 

[129348495](https://www.pivotaltracker.com/n/projects/1456570/stories/129348495)

### DIFF
--- a/infrastructure/fakes/fake_settings_source.go
+++ b/infrastructure/fakes/fake_settings_source.go
@@ -1,8 +1,6 @@
 package fakes
 
-import (
-	boshsettings "github.com/cloudfoundry/bosh-agent/settings"
-)
+import boshsettings "github.com/cloudfoundry/bosh-agent/settings"
 
 type FakeSettingsSource struct {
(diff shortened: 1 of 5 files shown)


DennisDenuto:
Reverting due to failing vsphere BATs tests.

Specifically, vsphere BATs test around waiting for monit to start up is timing out due to the agent bootstrap process taking 10X longer with these changes when fetching settings from CD-ROM (from 2.5 seconds to 25 seconds)

cppforlife:
@beyhan lets chnage this PR to work within http metadata fetching as it is the one that had problems connecting to metadata service.

cc @voelzmo

voelzmo:
@cppforlife already at it with https://www.pivotaltracker.com/story/show/133431439

Issue #107, opened 2016-12-05
Migrating data during disk size change is slow


CAFxX:
Changing disk size when the persistent volume contains a lot of data is very slow (e.g. migrating 600GB of data takes over 1 hour in our deployment)

bosh-agent already [uses](https://github.com/cloudfoundry/bosh-agent/blob/2ccba4ea4b75593572002b694ce68fb2d7d82058/platform/linux_platform.go#L1135), to speed up the process, two tar processes connected via pipe to migrate data from one disk to the other during disk size changes:
```
 (tar -C /var/vcap/store -cf - .) | (tar -C /var/vcap/store_migration_target -xpf -)
```

The default tar parameters are to use ~10KB blocks. This causes a high number of IOPS that (likely) limit throughput.

strace output of one of the tar processes:
```
read(0, "*****"..., 10240) = 10240
write(4, "*****"..., 10240) = 10240
read(0, "*****"..., 10240) = 10240
write(4, "*****"..., 10240) = 10240
read(0, "*****"..., 10240) = 10240
write(4, "*****"..., 10240) = 10240
```

output of `iostat 10`:
```
avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           0.23    0.00    4.08   17.44    0.00   78.25

Device:            tps    kB_read/s    kB_wrtn/s    kB_read    kB_wrtn
scd0              0.00         0.00         0.00          0          0
sda               1.80        13.60         9.20        136         92
sdb               0.00         0.00         0.00          0          0
sdc            1174.30    136634.40         0.00    1366344          0
sdd             353.30         4.40    179683.20         44    1796832
```

We propose to use the `tar` `-b` flag to use larger buffers and, if the pipe between the processes becomes the bottleneck, a `buffer` process between the two `tar`.

Using a 64K block size (the default pipe capacity on linux; this should already lower significantly the number of IOPS)
```
(tar -C /var/vcap/store  -b 128 -cf - .) | (tar -C /var/vcap/store_migration_target -b 128 -xpf -)
```

Using a 512K block size (with `buffer` in between to avoid blocking):
```
(tar -C /var/vcap/store -b 1024 -cf - .) | buffer | (tar -C /var/vcap/store_migration_target -b 1024 -xpf -)
```

CAFxX:
A (completely different) potential solution to this problem is in https://github.com/cloudfoundry/bosh/issues/1569

CAFxX:
[...]

PR #108, opened 2016-12-05
Move data in 64KB blocks when migrating persistent disks

CAFxX:
instead of using the default (10KB)

See #107


cppforlife:
nice change @CAFxX. will try to review with @tylerschultz tomorrow.

cppforlife:
@CAFxX out of curiosity what's the speed up for 600gb test you've done in your deployment?

CAFxX:
@cppforlife see https://github.com/cloudfoundry/bosh-agent/issues/107 for the details, copying data happens at ~150MB/s (generating >1000 IOPS especially on the read side), so copying 600GB take over one hour

CAFxX:
@cppforlife sorry upon further testing - making sure that the conditions are more realistic - it looks like this is not really giving much benefit. We're going to try some more and report back.

cppforlife:
Closing, for now.

Issue #85, opened 2016-08-03
Drop `delaycompress` from the logrotate configuration


CAFxX:
Since to maximize compatibility we're using `copytruncate` in the logrotate configuration it doesn't make much sense to [also use `delaycompress`](https://github.com/cloudfoundry/bosh-agent/blob/8a1a382bf63892f833935c1cf79afc4bdd3fb78b/platform/linux_platform.go#L474-L475). The latter is supposed to be used "[when some program cannot be told to close its logfile and thus might continue writing to the previous log file for some time](http://linux.die.net/man/8/logrotate)". This doesn't apply when logs are rotated with `copytruncate`.

The downside of using `delaycompress` is that a 1:1 copy of the current log file needs to be created (using temporarily twice the amount of disk space and causing more IO write load). Without `delaycompress` we would compress immediately, using less disk space and generating less IO write load.

CAFxX:
Can we send a PR for this?

CAFxX:
Included in stemcells 3363

PR #109, opened 2016-12-05
Do not delaycompress in logrotate

CAFxX:
It basically has only downsides when used in combination with copytruncate

See #85


CAFxX:
Any chance of this going in?

cppforlife:
@CAFxX will be in the upcoming stemcell update.

cppforlife:
@CAFxX i was just chatting with @tylerschultz and there is an interesting point about ease of grepping for recently rotated logs. wdyt about that use case?

CAFxX:
> wdyt about that use case?

I think that `zgrep` is your friend (as it works on both compressed and uncrompressed files) :)

tylerschultz:
@CAFxX ya, DK pointed that out right away. LGTM. Let's please merge into `develop` instead of master.

CAFxX:
Great, thanks!

CAFxX:
@tylerschultz @cppforlife ready for merge to develop

Issue #110, opened 2016-12-06
/tmp is not cleaned up during VM reboot


mattcui:
Due to the design change in [this commit](https://github.com/cloudfoundry/bosh-agent/commit/d466f198aa9d3f2c7269946d9798e1c8535cc2aa#diff-bc13c7242dfb8351e889278de03af265 ), the root_tmp dir is created with "mkdir -p", so if the dir root_tmp already exists, it will not be recreated, as a result, the /tmp is not cleaned up during VM reboot. Due to this change, our DEA (CF v235) can't work properly after VM reboot, since /tmp/warden dir is not removed. We know that this problem has already been fixed from CF release side, but I wonder if it's a right behavior to not clean up /tmp after VM reboot. Thanks.

cppforlife:
@mattcui i think it would be nice to clean up tmp though i dont think anyone should expect that /tmp is cleared after reboot. on some distributions (centos i think) /tmp isnt cleared.

mattcui:
@cppforlife So you agree to **recreate** root_tmp at the following line code, right? If so, we will submit a PR to address it.

https://github.com/cloudfoundry/bosh-agent/blob/d466f198aa9d3f2c7269946d9798e1c8535cc2aa/platform/linux_platform.go#L702

maximilien:
@mattcui closing this for now per our discussion. let's reopen if we find other software caring about this problem.

PR #111, opened 2016-12-07
Clean up root_tmp during bootstrap

mattcui:
For addressing [issue 110](https://github.com/cloudfoundry/bosh-agent/issues/110), clean up root_tmp before creating it during bootstrap.

/cc @maximilien

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -852,6 +852,11 @@ func (p linux) SetupTmpDir() error {
 		return nil
 	}
 
+	err = p.fs.RemoveAll(boshRootTmpPath)
+	if err != nil {
+		return bosherr.WrapError(err, fmt.Sprintf("cleaning up %s", boshRootTmpPath))
+	}
+
 	_, _, _, err = p.cmdRunner.RunCommand("mkdir", "-p", boshRootTmpPath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cppforlife:
@mattcui i dont think we can just use RemoveAll since if agent restart it will/try-to delete currently used files.

mattcui:
@cppforlife Do you have any good idea on how to distinguish between system reboot and bosh-agent restart. But before 3262.2, there created root_tmp through truncate command, it should refresh /tmp when restarting bosh-agent. Thanks.

cppforlife:
@mattcui i m not eager to add another tmpfs just to track this, may be /var/vcap/sys/run tmpfs can be reused but that seems way to hacky. why do you we need this again? it seems like any software that cant handle /tmp dir being in its prev state will have a similar problem if it just gets restarted.

mattcui:
@cppforlife I am not to intended to use truncated root_tmp again, I just meant before 3262.2, there always cleaned up /tmp during each agent restart, and you agreed to clean up /tmp during system reboot, so we need a way here to get this PR move forward, need your suggestions. Thanks.

cppforlife:
agent restart never caused /tmp cleanup. prev it was mounted as tmpfs i believe but was removed because we removed tmp space limitation.

maximilien:
@mattcui closing per issue comments.

PR #116, opened 2017-01-24
Configure networking on the virtual machine for SoftLayer

jianqiu:
To configure networking of the VM in SoftLayer is quite different with other IaaS, SoftLayer currently offers 2 different types of IP blocks, Static and Portable. For most cases, we are using Portable IPs to provision VM` networking. [https://knowledgelayer.softlayer.com/learning/virtual-machine-network-setup] is the reference we followed, in short, network configuration is done through a single configuration file and this file will need to be manually edited. In this PR, we implement bosh-agent to setup the networking by editing /etc/network/interfaces file. For example, 
```
Example public IP block – 192.0.2.0/29

·IP Address – 192.0.2.2
·Gateway – 192.0.2.1
·Subnet Mask – 255.255.255.248

Example private IP block – 10.0.0.0/29

·IP Address – 10.0.0.2
·Gateway – 10.0.0.1
·Subnet Mask – 255.255.255.248

Public and Private Network

Edit the following file with any text editor. Replace example IP addresses with the correct IP’s from your Public and Private IP blocks.

·/etc/network/interfaces
auto lo
iface lo inet loopback

auto eth1
iface eth1 inet static
address 192.0.2.2
netmask 255.255.255.248
gateway 192.0.2.1

auto eth0
iface eth0 inet static
address 10.0.0.2
netmask 255.255.255.248

#Static route for backend service network
up route add -net 10.0.0.0/8 gw 10.0.0.1
```

As the property `preconfigured` is no longer useful for SoftLayer, we get rid of it.
It includes two more properties in `network_settings`, `LinkName` and `Routes`.
About `Routes`,  it needs to setup persistent static route by SoftLayer, such as `setup persistent static route`, so we add this parameter to store this kind of static routes.
About `LinkName`, bosh-agent could detect mac address, and map interface name to network settings by mac address, but for SoftLayer, it is not entirely appropriate,  we use this property to indicate interface name, softlayer_cpi will propagate it through network_settings.

### DIFF
--- a/platform/net/ubuntu_net_manager.go
+++ b/platform/net/ubuntu_net_manager.go
@@ -86,9 +86,11 @@ func (net UbuntuNetManager) ComputeNetworkConfig(networks boshsettings.Networks)
 }
 
 func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, errCh chan error) error {
-	if networks.IsPreconfigured() {
-		// Note in this case IPs are not broadcasted
-		return net.writeResolvConf(networks)
+	if networks.HasLinkName() {
+		err := net.writeResolvConf(networks)
+		if err != nil {
+			return bosherr.WrapError(err, "Writing resolv configuration")
+		}
 	}
 
 	staticConfigs, dhcpConfigs, dnsServers, err := net.ComputeNetworkConfig(networks)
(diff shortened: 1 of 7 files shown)


platform/net/ubuntu_net_manager.go
```diff
@@ -86,9 +86,11 @@ func (net UbuntuNetManager) ComputeNetworkConfig(networks boshsettings.Networks)
 }
 
 func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, errCh chan error) error {
-	if networks.IsPreconfigured() {
```
cppforlife: preconfigured wasn't just added for softlayer; warden cpi for example uses it.

cppforlife:
> About LinkName, bosh-agent could detect mac address, and map interface name to network settings by mac address, but for SoftLayer, it is not entirely appropriate, we use this property to indicate interface name, softlayer_cpi will propagate it through network_settings.

why do you think it's appropriate?

cppforlife:
Closing this PR per discussion with @jianqiu. Waiting this to be reworked to work more generically.

PR #117, opened 2017-01-24
Registry settings fetcher retries on failure

beyhan:
This change requires changes introduced to bosh-utils: https://github.com/cloudfoundry/bosh-utils/pull/8. Please merge that one first, and bump bosh-utils in bosh-agent before merging this PR.

- make fetching from registry settings more
  resilient by using the RetryClient
- retry 10 times with 1s delay

[#135338075](https://www.pivotaltracker.com/story/show/135338075)

Signed-off-by: Beyhan Veli <beyhan.veli@sap.com>

### DIFF
--- a/infrastructure/complex_settings_source.go
+++ b/infrastructure/complex_settings_source.go
@@ -39,3 +39,7 @@ func (s ComplexSettingsSource) Settings() (boshsettings.Settings, error) {
 
 	return registry.GetSettings()
 }
+
+func (s ComplexSettingsSource) GetMetadataService() MetadataService {
+	return s.metadataService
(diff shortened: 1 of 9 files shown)


friegger:
@cppforlife We applied your suggested changes. We used `boshhttpclient.CreateDefaultClientInsecureSkipVerify()` for the inner http client, as no `certpool` was at hand.

Kiemes:
@cppforlife We applied the comments you made in https://github.com/cloudfoundry/bosh-utils/pull/8. We are closing the bosh-utils PR as it is not needed anymore.

PR #123, opened 2017-03-14
take desiredSwapSizeInBytes into account when partitioning root device

ampersand8:
We're currently working on a Kubernetes bosh-release. We noticed some warnings that with Kubernetes 1.6 swap will be a fatal error.

`W0308 08:27:30.075226 6465 container_manager_linux.go:205] Running with swap on is not supported, please disable swap! This will be a fatal error by default starting in K8s v1.6! In the meantime, you can opt-in to making this a fatal e`

Therefore we set "env.bosh.swap_size: 0" (stemcell 3363.12), but our bosh vms are still getting swap space. This is probably because the ephemeral path is not recognized and therefore we use [p.createEphemeralPartitionsOnRootDevice](https://github.com/cloudfoundry/bosh-agent/blob/master/platform/linux_platform.go#L581) instead of p.partitionEphemeralDisk.

@tjvman  Maybe there is a reason why desiredSwapSizeInBytes is not considered in createEphemeralPartitionsOnRootDevice but since I don't see why, I propose this PR.

The feature to define swap size was introduced with this commit [066ddb38409e217a06618643e223e9d4f30685fc](https://github.com/cloudfoundry/bosh-agent/commit/066ddb38409e217a06618643e223e9d4f30685fc).

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -578,7 +578,7 @@ func (p linux) SetupEphemeralDiskWithPath(realPath string, desiredSwapSizeInByte
 			return bosherr.Error("No ephemeral disk found, cannot use root partition as ephemeral disk")
 		}
 
-		swapPartitionPath, dataPartitionPath, err = p.createEphemeralPartitionsOnRootDevice()
+		swapPartitionPath, dataPartitionPath, err = p.createEphemeralPartitionsOnRootDevice(desiredSwapSizeInBytes)
 		if err != nil {
 			return bosherr.WrapError(err, "Creating ephemeral partitions on root device")
 		}
@@ -1310,7 +1310,7 @@ func (p linux) findRootDevicePathAndNumber() (string, int, error) {
 	return "", 0, bosherr.Error("Getting root partition device")
 }
 
-func (p linux) createEphemeralPartitionsOnRootDevice() (string, string, error) {
+func (p linux) createEphemeralPartitionsOnRootDevice(desiredSwapSizeInBytes *uint64) (string, string, error) {
 	p.logger.Info(logTag, "Creating swap & ephemeral partitions on root disk...")
 	p.logger.Debug(logTag, "Determining root device")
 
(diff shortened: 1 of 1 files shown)


cppforlife:
@ampersand8 will check it out.

cppforlife:
I just used 3363.12 to see. I dont have any swaps if swap_size is 0.

```
$ lsblk
NAME    MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
xvda    202:0    0   3G  0 disk
└─xvda1 202:1    0   3G  0 part /
xvdb    202:16   0  25G  0 disk
└─xvdb1 202:17   0  25G  0 part /var/vcap/data

$ cat /proc/swaps
Filename				Type		Size	Used	Priority
```

when used with following config:

```
- azs: [z1]
  instances: 1
  jobs: []
  name: zookeeper
  networks:
  - name: default
  stemcell: default
  vm_type: default
  env:
    bosh:
      swap_size: 0
```

ampersand8:
Hm, I'm also using 3363.12 and swap is always set.

Manifest extract:
```
- azs:
  - z1
  env:
    bosh:
      swap_size: 0
    persistent_disk_fs: xfs
  instances: 3
```

Swap partition is created on root disk:
```
$ lsblk
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
vda    253:0    0    80G  0 disk 
  vda1 253:1    0   2.9G  0 part /
  vda2 253:2    0   7.8G  0 part [SWAP]
  vda3 253:3    0  69.4G  0 part /var/vcap/data
vdb    253:16   0    64M  0 disk 
vdc    253:32   0    24G  0 disk 
  vdc1 253:33   0    24G  0 part /var/vcap/store

$ cat /proc/swaps 
Filename                                Type            Size    Used    Priority
/dev/vda2                               partition       8175112 3558072 -1
```

In the settings I see that swap_size is set to 0:
```
$ python -m json.tool /var/vcap/bosh/settings.json                                                                                                                                                                        
...
    "env": {
        "bosh": {
            "swap_size": 0
        },
        "persistent_disk_fs": "xfs"
    },
...
```

cppforlife:
@ampersand8 can you cat /var/vcap/bosh/etc/stemcell_version from inside your machine? may be it something to do with xfs but that would be surprising...

ampersand8:
`$ cat /var/vcap/bosh/etc/stemcell_version
3363.12`

Also see the bosh logs, which indicate that the realPath is actually resolved with mappedDevicePathResolver.GetRealDevicePath as we should otherwise see [another error here](https://github.com/cloudfoundry/bosh-agent/blob/master/infrastructure/devicepathresolver/virtio_device_path_resolver.go#L47). 

Here the bosh logs ([/var/vcap/bosh/log/@4000000058d4e6bb08c74e8c.s](https://gist.github.com/ampersand8/78a2fb30f9257e6df43c18984d6bd092)) from an affected vm.

cppforlife:
@ampersand8 not sure then, it definitely work in my test env. do you mind popping into #bosh channel on cloudfoundry slack. we can do some screen sharing.

cppforlife:
@ampersand8 ahh misunderstanding on my part. this doesnt work if ephemeral partition is on the root disk. do you mind adding a unit test that verifies this PR?

voelzmo:
fixed in https://github.com/cloudfoundry/bosh-agent/pull/140

PR #124, opened 2017-03-21
add useful testing script to quickly update stemcells with local agent binary

cppforlife:
cc @pivotal-jamil-shamy @tylerschultz @dpb587-pivotal

### DIFF
--- a/bin/repack-stemcell/README.md
+++ b/bin/repack-stemcell/README.md
@@ -1,0 +1,25 @@
+# Repacking Stemcell with Updated Agent
+
+!!! For debugging only !!!
+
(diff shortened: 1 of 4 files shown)


tylerschultz:
This looks fantastic.

Should this task folder be called something else? It's not actually building a stemcell, it's re-packing it with a modified agent.

cunnie:
@tylerschultz good point, I'll change the directory: `build-stemcell` &rarr; `repack-stemcell-with-new-agent` and re-submit PR (if that's okay with you, @cppforlife )

cppforlife:
@cunnie how about `repack-stemcell` to be concise.

cunnie:
@tylerschultz : directory (and documentation) have been modified to convey that we are _repacking_ an existing stemcell rather than _building_ a new stemcell. Merge at your convenience.

PR #125, opened 2017-03-23
Allow overriding source volume readahead during persistent disk migration

CAFxX:
This implements the workaround discussed in https://github.com/cloudfoundry/bosh-agent/issues/107#issuecomment-282908446. We have validated that raising readahead consistently gives a 3-4x improvement in migration throughput in our vSphere infrastructure.

Unit tests are passing and the approach has been validated, but we're currently unable to run the full bosh acceptance test suite.

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -80,6 +80,10 @@ type LinuxOptions struct {
 	// Strategy for resolving ephemeral & persistent disk partitioners;
 	// possible values: parted, "" (default is sfdisk if disk < 2TB, parted otherwise)
 	PartitionerType string
+
+	// When set to > 0 enables overriding the defaut readahead for the duration of
+	// the persistent volume migration. If set to <=0 disables the override.
+	PersistentDiskMigrationReadahead int
 }
 
 type linux struct {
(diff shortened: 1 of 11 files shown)


cppforlife:
@CAFxX im a bit conservative about changes to persistent disk handling, do you mind adding an agent option (`env.bosh.persistent_disk_migration_readahead`?) that would enable this behaviour to roll this out more gradually?

CAFxX:
@cppforlife made it configurable, but I now need some help about how to propagate the configuration from the bosh spec to the agent.

CAFxX:
Just for the record, the change is pretty safe: readahead is already on by default with a value of 256; this PR only raises it to a slightly higher number for the duration of the migration. It is effectively performance for free.

CAFxX:
Added an extra unit test

CAFxX:
@cppforlife gentle ping

cppforlife:
@CAFxX i was referring to following settings/env (https://github.com/cloudfoundry/bosh-agent/blob/master/settings/settings.go#L161). that way your deployment manifest can just specify it (no need to change director configuration).

CAFxX:
@cppforlife sorry for the delay, simplified logic and made configurable via boshenv

CAFxX:
@cppforlife gentle ping

CAFxX:
@cppforlife gentle ping

cppforlife:
@CAFxX oops, lost this one. will review tomorrow with @tylerschultz and try to merge in for upcoming stemcell.

CAFxX:
Great, thanks!

platform/linux_platform.go
```diff
@@ -1176,6 +1219,11 @@ func (p linux) MigratePersistentDisk(fromMountPoint, toMountPoint string) (err e
 	return
 }
 
+func (p linux) SetPersistentDiskMigrationReadahead(readahead int) error {
+	p.options.PersistentDiskMigrationReadahead = readahead
```
cppforlife: i dont think this will properly work since readahead will be assigned to a copy of p.options (p isnt a pointer). have you seen it work on a live machine?

platform/linux_platform_test.go
```diff
@@ -2700,6 +2700,38 @@ Number  Start   End     Size    File system  Name             Flags
...
+			mounter.IsMountPointErr = nil
+			cmdRunner.AddCmdResult(
+				"blockdev --getra /dev/sdf",
+				fakesys.FakeCmdResult{Stdout: "256"},
+			)
+			options.PersistentDiskMigrationReadahead = 1024
+		})
+
+		It("migrates persistent disk with readahead override", func() {
+			err := platform.MigratePersistentDisk("/from/path", "/to/path")
```
cppforlife: you can use `SetPersistentDiskMigrationReadahead` for this test to see that p.options isnt actually persistented after modification.

jaresty:
Merge story [#152974652](https://www.pivotaltracker.com/story/show/152974652)

PR #129, opened 2017-05-12
allow enabling of ipv6 through env.bosh.ipv6.enable=true

cppforlife:
Users can force Ubuntu machines to enable IPv6 at the kernel level through agent setting:

```yaml
instance_groups:
- name: my-stuff
  env:
    bosh:
      ipv6:
        enable: true
```


cppforlife:
Related PR for e2e test: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/6

PR #131, opened 2017-06-19
Add vm uptime information into vitals

barthy1:
vitals.uptime.secs field is added as expected by bosh-cli

[#146195069](https://www.pivotaltracker.com/story/show/146195069)

### DIFF
--- a/platform/stats/stats_collector_interface.go
+++ b/platform/stats/stats_collector_interface.go
@@ -26,6 +26,14 @@ type DiskStats struct {
 	InodeUsage Usage
 }
 
+type Seconds struct {
+	Seconds uint64
+}
+
+type VMStats struct {
+	Uptime Seconds
+}
+
 type Collector interface {
 	StartCollecting(time.Duration, chan struct{})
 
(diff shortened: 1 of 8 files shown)


platform/stats/stats_collector_interface.go
```diff
@@ -35,6 +43,8 @@ type Collector interface {
 	GetMemStats() (usage Usage, err error)
 	GetSwapStats() (usage Usage, err error)
 	GetDiskStats(mountedPath string) (stats DiskStats, err error)
+
+	GetVMStats() (stats VMStats, err error)
```
dpb587-pivotal: Hi (and sorry for the delay) again - curious if you recall a reason why `GetVMStats` was used, rather than something like `GetUptimeStats`? Seems like CPU/Mem/Swap/Disk are also VM-specific stats as well, and in theory, either they should be moved into `VMStats`, or the function should be renamed to `GetUptimeStats`. If so, then maybe we can eliminate one of the extra `Seconds` or `VMStats` structs, too. We could instead use one `UptimeStats` with `Seconds` as a field.

Was curious if there's more to this that we're not seeing. We're happy to make minor changes like this, since it's been a while. Thanks!

barthy1:
@dpb587-pivotal answering to your question about VMStats - I thought it'd be better to have some common struct for all single vm params, but yeah, now I understand, that disk data, memory data, etc also can be considered as VMStats. So, I've updated the code to have UptimeStats

dpb587-pivotal:
g2g :shipit:

PR #141, opened 2017-09-20
ipv6 for manual networking on ubuntu

cppforlife:
- tested on vsphere with ipv6 director + zookeeper deployment (local dns enabled to make sure zookeeper isnt confused by colons)


zachgersh:
@cppforlife got a conflict here - gonna need to resolve that.

cppforlife:
conflict resolved in https://github.com/cloudfoundry/bosh-agent/pull/143

dpb587-pivotal:
@cppforlife, you should reuse pull requests, especially to retain prior discussions/context

PR #142, opened 2017-09-29
Add preconfigured network support to opensuse net manager

manno:
This is needed for bosh-lite/warden stemcell networking


manno:
@dpb587-pivotal I think you briefly looked at the PR while we talked at the summit. I would just rebase, test and merge this if that's ok, since it would only affect openSUSE?

PR #147, opened 2017-11-20
Set correct permissions (755) ephermeral mountdir

sleepsonthefloor:
* Also enforce that these permissions are correct on the
   mounted file system
 * This allows the bosh-agent to start when SkipDiskSetup=true

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	ephemeralDiskPermissions  = os.FileMode(0750)
+	ephemeralDiskPermissions  = os.FileMode(0755)
 	persistentDiskPermissions = os.FileMode(0700)
 
 	logDirPermissions         = os.FileMode(0750)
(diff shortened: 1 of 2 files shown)


dpb587-pivotal:
I believe this is the pull request following up to the conversation in [cloudfoundry#oracle](https://cloudfoundry.slack.com/archives/C7P97LNFL/p1510701010000038) where you noticed that non-mounted ephemeral disks did not have the same permissions as mounted ephemeral disks; and this is trying to correct that difference.

Merge story: https://www.pivotaltracker.com/story/show/153101052

Updated the branch from `master` to `develop`.

dmutreja:
Hi @dpb587-pivotal any word on when this can be merged? 

We are looking to submit a PR to https://github.com/oracle/bosh-linux-stemcell-builder and would like the remove the [workaround for this issue](https://github.com/oracle/bosh-linux-stemcell-builder/blob/oracle_stemcell_base/stemcell_builder/stages/system_oracle/apply.sh#L22-L23)  prior to submitting that PR.

Thanks!
/cc @cppforlife

dpb587-pivotal:
Thanks for the reminder.

@cppforlife please prioritize if it's not already; [story](https://www.pivotaltracker.com/story/show/153101052) already exists.

PR #148, opened 2017-12-04
Pre-PR for initial feedback: Support network alias

maximilien:
Do not merge this. This is to allow me to make comments to @EdwardStudy


platform/net/interface_configuration_creator_test.go
```diff
@@ -151,6 +151,34 @@ func describeInterfaceConfigurationCreator() {
 					})
 				})
 			})
+
+			Context("And the network has a alias", func() {
```
maximilien: Minor: "[...] has a alias" => "[...] has an alias"

platform/net/interface_configuration_creator_test.go
```diff
@@ -151,6 +151,34 @@ func describeInterfaceConfigurationCreator() {
...
+				BeforeEach(func() {
+					staticNetwork.Alias = "static-interface-name"
```
maximilien: change "static-interface-name" => "static-interface-alias" or similar so it's clear from name

platform/net/interface_configuration_creator_test.go
```diff
@@ -151,6 +151,34 @@ func describeInterfaceConfigurationCreator() {
...
+					interfacesByMAC["fake-static-mac-address"] = "any-interface-name"
+				})
+
+				It("creates an interface configuration when matching interface exists", func() {
+					staticInterfaceConfigurations, dhcpInterfaceConfigurations, err := interfaceConfigurationCreator.CreateInterfaceConfigurations(networks, interfacesByMAC)
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(staticInterfaceConfigurations).To(Equal([]StaticInterfaceConfiguration{
+						StaticInterfaceConfiguration{
+							Name:                "any-interface-name",
```
maximilien: This would then change to name above with "alias"

platform/net/interface_configuration_creator_test.go
```diff
@@ -239,6 +267,47 @@ func describeInterfaceConfigurationCreator() {
 					})
 				})
 			})
+
+			Context("when static network has alias, dhcp network is not allowd to have alias", func() {
```
maximilien: Can you include some links to the reason for this? I imagine there is a IETF RFC discussing this. Just add a link to section for this rule. Thx.

platform/net/ubuntu_net_manager.go
```diff
@@ -102,6 +102,11 @@ func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, errC
 		return net.writeResolvConf(networks)
 	}
 
+	// Specific for softlayer networking
```
maximilien: Cool, great comment. Though let's be consistent while not adding this everywhere. Perhaps first time SL-specific code shows up. Thx

platform/net/ubuntu_net_manager.go
```diff
@@ -137,7 +142,20 @@ func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, errC
 
 	staticAddresses, dynamicAddresses := net.ifaceAddresses(staticConfigs, dhcpConfigs)
 
-	err = net.interfaceAddressesValidator.Validate(staticAddresses)
+	// Skip the validation of virtual interfaces(like eth0:0, eth0:1... )
+	staticAddressesWithoutVirtual := []boship.InterfaceAddress{}
+	r, err := regexp.Compile(`:\d+`)
+	if err != nil {
+		return bosherr.WrapError(err, "There is a problem with your regexp.")
```
maximilien: Be specific about which Regexp that is failing, e.g., `"There is a problem with your regexp: `:\d+` that is used to skip validation of virtual interfaces, e.g., eth0:0, eth0:1, ..."` With that you may even be able to remove comment above :)

platform/net/ubuntu_net_manager_test.go
```diff
@@ -904,6 +938,108 @@ iface ethstatic inet static
 				Expect(networkConfig.StringContents()).To(Equal(expectedNetworkConfiguration))
 			})
 		})
+
+		Context("when manual networks were configured with portal IP", func() {
```
maximilien: I don't understand the title for this context? What portal IP are we tabling about? I am guessing you mean `portable` IPs?

platform/net/ubuntu_net_manager_test.go
```diff
@@ -904,6 +938,108 @@ iface ethstatic inet static
...
+					boship.NewSimpleInterfaceAddress("eth1", "169.50.68.75"),
+				}
+				fs.WriteFileString("/etc/resolv.conf", `
+nameserver 8.8.8.8
+nameserver 10.0.80.11
+nameserver 10.0.80.12
+`)
+			})
+
+			It("succeeds", func() {
```
maximilien: What happens when it fails? There are a few error conditions to consider? We should have tests for these, no?

settings/settings.go
```diff
@@ -334,6 +336,21 @@ func (n Networks) IPs() (ips []string) {
 	return
 }
 
+func (n Networks) HasInterfaceAlias() bool {
```
maximilien: Since this function is public, you need to have specific tests for it. So either you make private and it's tested in code that uses it or you have tests for it. If you keep public you also likely need to promote it in the `Networks` interface definition and make sure all struct implementing this (might only be this one) have suitable implementation and tests.

maximilien:
@EdwardStudy please update the PR when you have made necessary changes and comment here. I am not sure which of the feedback is addressed.

EdwardStudy:
@maximilien We should close it and re-open a formal PR about this feature.

Thank you~

maximilien:
Closing to @EdwardStudu can resubmit with fixes

PR #150, opened 2017-12-04
Pre-PR for initial feedback: add iscisi device path resolver for softlayer

maximilien:
Do not merge this. This is to allow me to make comments to @EdwardStudy


platform/openiscsi/concrete_open_iscsi_admin.go
```diff
@@ -0,0 +1,162 @@
...
+node.conn[0].timeo.noop_out_timeout = 15
+node.session.iscsi.InitialR2T = No
+node.session.iscsi.ImmediateData = Yes
+node.session.iscsi.FirstBurstLength = 262144
+node.session.iscsi.MaxBurstLength = 16776192
+node.conn[0].iscsi.MaxRecvDataSegmentLength = 65536
+`
+
+func (iscsi ConcreteOpenIscsiAdmin) Start() (err error) {
+	iscsi.logger.Debug(iscsi.logtag, "Start Open-iscsi deamon")
```
maximilien: Look at how other BOSH code and see how Debug vs Info is used. This looks like tracing info

platform/openiscsi/concrete_open_iscsi_admin.go
```diff
@@ -0,0 +1,162 @@
...
+	_, _, _, err = iscsi.runner.RunCommand("/etc/init.d/open-iscsi", "start")
+	return
+}
+
+func (iscsi ConcreteOpenIscsiAdmin) Stop() (err error) {
+	iscsi.logger.Debug(iscsi.logtag, "Stop Open-iscsi deamon")
+	_, _, _, err = iscsi.runner.RunCommand("/etc/init.d/open-iscsi", "stop")
```
maximilien: In all these methods, what if there is an error? You are ignoring that and all other return values. Seems like this code is incomplete.

Also, I'd expect unit tests for each.

infrastructure/devicepathresolver/iscsi_device_path_resolver.go
```diff
@@ -0,0 +1,175 @@
...
+		contents, err := ispr.fs.ReadFile(managedDiskSettingsPath)
+		if err != nil {
+			return "", bosherr.WrapError(err, "Reading managed_disk_settings.json")
+		}
+		lastMountedCid = string(contents)
+
+		return lastMountedCid, nil
+	}
+
+	return "", nil
```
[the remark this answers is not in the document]
EdwardStudy: I thought is not occur an error. The file(managed_disk_settings.json) would be written to disk cid after mount iSCSI disk. If the file does not exist, The CID will be "".

platform/disk/parted_partitioner.go
```diff
@@ -309,6 +309,11 @@ func (p partedPartitioner) createMapperPartition(devicePath string) error {
 		return bosherr.WrapError(err, "Shelling out to restart open-iscsi")
 	}
 
+	_, _, _, err = p.cmdRunner.RunCommand("/etc/init.d/multipath-tools", "restart")
```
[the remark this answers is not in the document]
EdwardStudy: It would returns error if the command runs and exits with a non-zero exit status.

platform/openiscsi/concrete_open_iscsi_admin.go
EdwardStudy: Yes, these methods only return error directly. Because the cmd runner in bosh-utils could write log file entries about the running command stdout/stderr. I have referred to other code in bosh-agent. If these commands' outputs do not participate in the follow-up attendance, ignore these outputs.

EdwardStudy:
I had finished adding test cases and fixing mistakes.
Some replies had also be added.

@maximilien Please help to review, thanks.

maximilien:
@EdwardStudy please update the PR when you have made necessary changes and comment here. I am not sure which of the feedback is addressed.

EdwardStudy:
@maximilien We should close it and re-open a formal PR about this feature.

maximilien:
Closing to @edwardstudy can resubmit with fixes

PR #155, opened 2018-01-04
Support for PhotonOS 2.0

DheerajSShetty:
- A new platform called PhotonOS
- Used netmgr which is a Network Manager for Photon OS.


DheerajSShetty:
Will close this PR for now. Requires some CLA work before this can be submitted again.

PR #159, opened 2018-02-10
Add Photon OS support

DheerajSShetty:
- Support for Linux based platform PhotonOS

- Make netmgr calls to manage the Photon OS network

Signed-off-by: Dheeraj Shetty dheerajs@vmware.com

### DIFF
--- a/platform/cert/cert_manager.go
+++ b/platform/cert/cert_manager.go
@@ -83,6 +83,18 @@ func NewOpensuseOSCertManager(fs boshsys.FileSystem, runner boshsys.CmdRunner, t
 	}
 }
 
+func NewPhotonOSCertManager(fs boshsys.FileSystem, runner boshsys.CmdRunner, timeout time.Duration, logger logger.Logger) Manager {
+	return &certManager{
+		fs:            fs,
+		runner:        runner,
+		path:          "/etc/ssl/certs/",
+		updateCmdPath: "/usr/bin/rehash_ca_certificates.sh",
+		logger:        logger,
+		logTag:        "PhotonOSCertManager",
+		updateTimeout: timeout,
+	}
+}
+
 func NewDummyCertManager(fs boshsys.FileSystem, runner boshsys.CmdRunner, timeout time.Duration, logger logger.Logger) Manager {
 	return &certManager{
 		fs:            fs,
--- a/platform/cert/cert_manager_test.go
+++ b/platform/cert/cert_manager_test.go
@@ -385,6 +385,35 @@ var _ = Describe("Certificate Management", func() {
 			})
 		})
 
+		Context("PhotonOS", func() {
+			BeforeEach(func() {
+				fakeFs = fakesys.NewFakeFileSystem()
+				fakeCmdRunner = fakesys.NewFakeCmdRunner()
+				fakeCmdRunner.AddCmdResult("/usr/bin/rehash_ca_certificates.sh", fakesys.FakeCmdResult{
+					Stdout:     "",
+					Stderr:     "",
+					ExitStatus: 0,
+					Sticky:     true,
+				})
+				certManager = cert.NewPhotonOSCertManager(fakeFs, fakeCmdRunner, 0, log)
+			})
+
+			SharedLinuxCertManagerExamples("/etc/ssl/certs", "/usr/bin/rehash_ca_certificates.sh")
+
+			It("executes update cert command", func() {
+				fakeCmdRunner = fakesys.NewFakeCmdRunner()
+				fakeCmdRunner.AddCmdResult("/usr/bin/rehash_ca_certificates.sh", fakesys.FakeCmdResult{
+					Stdout:     "",
+					Stderr:     "",
+					ExitStatus: 2,
+					Error:      errors.New("command failed"),
+				})
+				certManager = cert.NewPhotonOSCertManager(fakeFs, fakeCmdRunner, 0, log)
+
+				err := certManager.UpdateCertificates(cert1)
+				Expect(err).To(HaveOccurred())
+			})
+		})
 		Context("Windows", func() {
 			const validCerts string = `-----BEGIN CERTIFICATE-----
 MIIC0jCCAboCCQCuQJScK+G0WzANBgkqhkiG9w0BAQsFADArMQswCQYDVQQGEwJV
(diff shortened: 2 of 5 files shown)


DheerajSShetty:
@cppforlife Added the tests for the code changes

DheerajSShetty:
@cppforlife Please let me know if there is anything that I need to do from my side.

platform/net/photonos_net_manager.go
```diff
@@ -0,0 +1,233 @@
+// Copyright © 2018 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache License 2.0
```
luan: Could you remove this copyright assignment from here? Copyrights are assigned to the CF Foundation on the NOTICE file at the root of the repo.

DheerajSShetty: Hi Luan,
May I know where should I add these CopyRights? Does this mean we should not write the VMware Copyright in this repo at all?

cppforlife:
> Does this mean we should not write the VMware Copyright in this repo at
all?

correct, all of this code is copyrighted by CFF (
https://github.com/cloudfoundry/bosh-agent/blob/master/NOTICE#L1)

DheerajSShetty:
I will remove the Copyright

DheerajSShetty:
@cppforlife @luan I have removed the CopyRights from the file

platform/cert/cert_manager.go
```diff
@@ -160,6 +173,13 @@ func (c *certManager) UpdateCertificates(certs string) error {
 		return bosherr.WrapError(err, "Running command to update certificates without retries")
 	}
 
+	if strings.Compare(c.logTag, "PhotonOSCertManager") == 0 {
```
jfmyers9: Can you please change this to `c.logTag == "PhotonOSCertManager"` since it is clearer and more performant; however, we're wondering if the `c.logTag` is the best way to detect we are on PhotonOS here -- can you please explain that decision?

platform/cert/cert_manager.go
```diff
@@ -160,6 +173,13 @@ func (c *certManager) UpdateCertificates(certs string) error {
...
+		_, _, _, err = c.runner.RunCommand("cat", fmt.Sprintf("%s*.pem > /etc/pki/tls/certs/ca-bundle.crt", c.path))
```
jfmyers9: Can we do this natively in golang instead of shelling out to `cat`?   [`io.Copy`](https://golang.org/pkg/io/#Copy) is a native way to do what you are doing here.

platform/net/photonos_net_manager.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return bosherr.WrapError(err, "Validating dns configuration")
+	}
+
+	net.broadcastIps(append(staticAddresses, dynamicAddresses...), errCh)
+
+	return nil
+}
+
+func (net photonosNetManager) GetConfiguredNetworkInterfaces() ([]string, error) {
```
jfmyers9: Can you write some tests for this method?

platform/cert/cert_manager.go
DheerajSShetty: Inside the update certificate function call, I found this a rather simply way to distinctly identify the platform. Also PhotonOSCertManager logtag wont be used by other Distribution.
Please correct me if I am wrong.

dpb587-pivotal:
🔗 Pivotal Tracker, [Story #155951359](https://www.pivotaltracker.com/story/show/155951359)

platform/cert/cert_manager.go
jaresty: We'd like to propose an alternative approach:
1. Write a script that calls `/usr/bin/c_rehash` and cats all the ca certificates into the `/etc/pki/tls/certs/ca-bundle.crt` file
2. Use that script as the `updateCmdPath` argument.
3. Remove the now-unnecessary command and logTag-testing conditional since the script has taken care of everything necessary for setting up PhotonOS.

jfmyers9: How would that script be distributed? The `updateCmdPath` seems to be primarily used for binaries that are built into the OS to perform CA certificate updates.

platform/net/photonos_net_manager.go
DheerajSShetty: Done.

DheerajSShetty:
Certificate manager calls a script managed by PhotonOS. Added tests for the method GetConfiguredNetworkInterfaces()

platform/cert/cert_manager.go
jrussett: Preferably we would like to have a PhotonOS stemcell native script that includes the ca certificate `cat` command, e.g. /usr/bin/photonos_cert_updater. This way we can maintain the old certmanager update certificate logic and simply call a PhotonOs-specific updateCmdPath. 

This script should be baked in the Photon OS stemcell.

platform/cert/cert_manager_test.go
```diff
@@ -385,6 +385,35 @@ var _ = Describe("Certificate Management", func() {
...
+					Stderr:     "",
+					ExitStatus: 0,
+					Sticky:     true,
+				})
+				certManager = cert.NewPhotonOSCertManager(fakeFs, fakeCmdRunner, 0, log)
+			})
+
+			SharedLinuxCertManagerExamples("/etc/ssl/certs", "/usr/bin/rehash_ca_certificates.sh")
+
+			It("executes update cert command", func() {
```
jfmyers9: This test doesn't pass when we ran it locally. Can you fix this?

DheerajSShetty: Fixed.

DheerajSShetty:
Squashed changes

DheerajSShetty:
I have added a script to PhotonOS which does certificate management. Also made sure the tests pass.
I have squashed changes and named it "Add PhotonOS support"
Please let me know if there is anything to be changed.

platform/net/photonos_net_manager.go
```diff
@@ -0,0 +1,230 @@
...
+	interfaces := []string{}
+
+	interfacesByMacAddress, err := net.detectMacAddresses()
+	if err != nil {
```
jrussett: Can you add a test for an error raised by `detectMacAddresses()`? It'd be great if you can test the same for `SetupNetworking()` which indirectly also uses `detectMacAddresses`.

christarazi:
Closing due to inactivity. Feel free to open again.

PR #164, opened 2018-03-28
Add ipv6 support for openSUSE

manno:
IPv6 was necessary to pass stemcell acceptance tests. We also fixed the unit tests and a problem with the way DNS servers were handled.


cppforlife:
@manno i m a bit surprised that this is necessary.

> IPv6 was necessary to pass stemcell acceptance tests

what were failures without this change?

aashah:
We've recently decided to move away from supporting more than a few build targets for the bosh-agent [1]. For that reason, I'm gonna close this issue as 'will not implement'.

Let me know if there's more here. I'm sorry there wasn't more discussion before closing.

[1] https://lists.cloudfoundry.org/g/cf-bosh/topic/upcoming_changes_to/65618897?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,65618897

PR #165, opened 2018-04-06
Setup /var/log/lastlog

aduffeck:
/var/log/lastlog is not written in openSUSE if it doesn't already exist.

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -985,6 +985,22 @@ func (p linux) SetupLogDir() error {
 		return bosherr.WrapError(err, "Chmoding sysstat log dir")
 	}
 
+	lastlogPath := path.Join(boshRootLogPath, "lastlog")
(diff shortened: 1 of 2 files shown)


cppforlife:
@aduffeck would you mind adding unit tests that assert that these command are called.

cppforlife:
@aduffeck brain fart from my side... can you switch to using fs's Chown/Chmod [1] and WriteFileString [2]

[1] https://github.com/cloudfoundry/bosh-utils/blob/87d3668a74af6d395aae97a3764aac279abbaf42/system/file_system_interface.go#L28-L29
[2] https://github.com/cloudfoundry/bosh-utils/blob/87d3668a74af6d395aae97a3764aac279abbaf42/system/file_system_interface.go#L33

cppforlife:
prioritized to pull in: https://www.pivotaltracker.com/story/show/157071295

christarazi:
Closing due to inactivity. Feel free to open again if you want to.

PR #175, opened 2018-06-19
Fix Linux hosts file formatting

mlmitch:
[This](https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution) is the Debian documentation I reference in the commit.

Debian documentation says that the hostname entry should be on a
seperate line from `localhost`.

Having it on the same line causes queries for fully qualified domain
names (e.g. `hostname -f`) to return "localhost".

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -421,7 +421,8 @@ func (p linux) SetupRecordsJSONPermission(path string) error {
 	return nil
 }
 
-const EtcHostsTemplate = `127.0.0.1 localhost {{ . }}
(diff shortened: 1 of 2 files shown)


mlmitch:
CLA has been submitted

mlmitch:
Refer [this issue](https://github.com/cloudfoundry/bosh-dns-release/issues/30) for some background. This PR only solves one of the things I bring up there.

Lafunamor:
seems that the CI job is stuck. Any chance this can be merged? This would help us a lot.

Lafunamor:
can we please revisit this one? It's still not merged...

Issue #184, opened 2018-08-01
Increase memory causes bosh agent fails to start


edwardstudy:
Hi, we are doing test for softlayer-cpi with xenial stemcell(agent v2.123.0). 

We just increased cell memory from 32G to 48G which caused the ephemeral disk was re-patitioned and formatted. Trusty stemcell doesn't have this problem.

For Xenial stemcell, after os reload, the memory is successfully modified (seen from SL portal) but bosh agent failed to start:

```
2018-08-01_10:02:27.42200 [linuxPlatform] 2018/08/01 10:02:27 INFO - Partitioning `/dev/xvdc' with [[Type: swap, SizeInBytes: 50636873728] [Type: linux, SizeInBytes: 271485673472]]
2018-08-01_10:02:27.42203 [Cmd Runner] 2018/08/01 10:02:27 DEBUG - Running command 'parted -m /dev/xvdc unit B print'
2018-08-01_10:02:27.43166 [Cmd Runner] 2018/08/01 10:02:27 DEBUG - Stdout: BYT;
2018-08-01_10:02:27.43167 /dev/xvdc:322122547200B:xvd:512:512:gpt:Xen Virtual Block Device:;
2018-08-01_10:02:27.43168 1:1048576B:33726398463B:33725349888B:linux-swap(v1):bosh-partition-0:;
2018-08-01_10:02:27.43168 2:33726398464B:322121498623B:288395100160B:ext4:bosh-partition-1:;
2018-08-01_10:02:27.43172 [Cmd Runner] 2018/08/01 10:02:27 DEBUG - Stderr:
2018-08-01_10:02:27.43173 [Cmd Runner] 2018/08/01 10:02:27 DEBUG - Successful: true (0)
2018-08-01_10:02:27.43175 [main] 2018/08/01 10:02:27 ERROR - App setup Running bootstrap: Setting up ephemeral disk: Partitioning ephemeral disk: Partitioning ephemeral disk '/dev/xvdc': '/dev/xvdc' contains a partition created by bosh. No partitioning is allowed.
2018-08-01_10:02:27.43176 [main] 2018/08/01 10:02:27 ERROR - Agent exited with error: Running bootstrap: Setting up ephemeral disk: Partitioning ephemeral disk: Partitioning ephemeral disk '/dev/xvdc': '/dev/xvdc' contains a partition created by bosh. No partitioning is allowed.
```

[...]

PR #186, opened 2018-08-16
Repartition ephemeral disk when OS reload

edwardstudy:
Hi,

This PR focused on repartition ephemeral disk and was originated from https://github.com/cloudfoundry/bosh-agent/issues/184

1. When OS reload, agent will repartition the ephemeral disk and set `ephemeral_disk_partitioned` to true.

2. When OS reboot or agent restart, agent will check `ephemeral_disk_partitioned` field and skip repartitioning if it is true.

/cc @mattcui @dpb587  Please help to review

### DIFF
--- a/platform/bootstrap_state.go
+++ b/platform/bootstrap_state.go
@@ -14,7 +14,8 @@ type BootstrapState struct {
 }
 
 type LinuxState struct {
-	HostsConfigured bool `json:"hosts_configured"`
+	HostsConfigured          bool `json:"hosts_configured"`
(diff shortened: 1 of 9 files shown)


edwardstudy:
Reopen after implementing  `EphemeralDevicePartitioner`: https://github.com/cloudfoundry/bosh-agent/issues/184#issuecomment-416772112

edwardstudy:
Reopen a PR: https://github.com/cloudfoundry/bosh-agent/pull/190

PR #187, opened 2018-08-28
Add feature flag to mount jobs directory on tmpfs

xoebus:
**Please do not merge this commit!**

This change allows operators to mount their /var/vcap/data/jobs directory (and /var/vcap/jobs, transitively) on a tmpfs to avoid the contents being written to disk.

The configuration lives under the env.bosh instance group manifest property:

```yaml
- name: redis
  ...
  env:
    bosh:
      job_dir:
        tmpfs: true
        tmpfs_size: 256m
```

This feature flag default to disabled and, if not set, the size of the tmpfs defaults to 100 megabytes. The size is configurable in the manifest to provide an interim escape-hatch if the default is too small for a job which stores a large files in its job directory (this should be uncommon).

The `FileBundle` change is required in order to avoid invalid cross-device link errors when trying to rename a job directory. The related tests are pended for the moment because they were very coupled to the previous implementation. I'll update them if we're happy with this approach.

---

There are still some things missing overall:

- [x] Is there a better format for passing the size through to the mount call? It's string-ly typed at the moment.
- [ ] Does this need an integration test? If so, where should it be added?
- [ ] Does this need an implementation on Windows? Can this be implemented on Windows?
- [ ] I needed to modify the `FileBundle` to avoid cross device rename issues. Is this safe? Is there a better way to do this?
- [ ] The BOSH agent [downloads the rendered jobs to a temporary directory first](https://github.com/cloudfoundry/bosh-agent/blob/cbfc5060b5b3955aa72c044a557ee734122f23b0/agent/applier/jobs/rendered_job_applier.go#L105). Is that mounted on tmpfs or would that need to be changed too?

I think I managed to alter an existing integration test to exercise this and there were no errors (I'm not sure I did it correctly though).

What do you think?

### DIFF
--- a/agent/applier/bundlecollection/file_bundle.go
+++ b/agent/applier/bundlecollection/file_bundle.go
@@ -66,9 +66,15 @@ func (b FileBundle) Install(sourcePath string) (boshsys.FileSystem, string, erro
 
 	// Rename MUST be the last possibly-failing operation
 	// because IsInstalled() relies on installPath presence.
-	err = b.fs.Rename(sourcePath, b.installPath)
+	err = b.fs.CopyDir(sourcePath, b.installPath)
 	if err != nil {
 		return nil, "", bosherr.WrapError(err, "Moving to installation directory")
+	}
+
+	// XXX(cbrown): disregard the comment above and profit
+	err = b.fs.Chown(b.installPath, "root:vcap")
+	if err != nil {
+		return nil, "", bosherr.WrapError(err, "Setting ownership on installation directory")
 	}
 
 	return b.fs, b.installPath, nil
--- a/agent/applier/bundlecollection/file_bundle_test.go
+++ b/agent/applier/bundlecollection/file_bundle_test.go
@@ -15,7 +15,7 @@ import (
 
 //go:generate counterfeiter -o fakes/fake_clock.go /code.cloudfoundry.org/clock Clock
 
-var _ = Describe("FileBundle", func() {
+var _ = XDescribe("FileBundle", func() {
 	var (
 		fs          *fakesys.FakeFileSystem
 		fakeClock   *fakes.FakeClock
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -127,7 +127,7 @@ func (boot bootstrap) Run() (err error) {
 		return bosherr.WrapError(err, "Starting up logging and auditing utilities")
 	}
 
-	if err = boot.platform.SetupDataDir(); err != nil {
+	if err = boot.platform.SetupDataDir(settings.Env.Bosh.JobDir); err != nil {
 		return bosherr.WrapError(err, "Setting up data dir")
 	}
 
--- a/agent/bootstrap_test.go
+++ b/agent/bootstrap_test.go
@@ -305,10 +305,26 @@ var _ = Describe("bootstrap", func() {
 			})
 		})
 
-		It("sets up data dir", func() {
-			err := bootstrap()
-			Expect(err).NotTo(HaveOccurred())
-			Expect(platform.SetupDataDirCallCount()).To(Equal(1))
+		Describe("setting up the data dir", func() {
+			It("sets up data dir", func() {
+				err := bootstrap()
+				Expect(err).NotTo(HaveOccurred())
+				Expect(platform.SetupDataDirCallCount()).To(Equal(1))
+			})
+
+			Context("when there are job directory specific feature flags", func() {
+				It("passes those through to the platform", func() {
+					settingsService.Settings.Env.Bosh.JobDir = boshsettings.JobDir{
+						TmpFs:     true,
+						TmpFsSize: "100M",
+					}
+
+					err := bootstrap()
+					Expect(err).NotTo(HaveOccurred())
+					Expect(platform.SetupDataDirCallCount()).To(Equal(1))
+					Expect(platform.SetupDataDirArgsForCall(0)).To(Equal(boshsettings.JobDir{TmpFs: true, TmpFsSize: "100M"}))
+				})
+			})
 		})
 
 		Context("when setting up the data dir fails", func() {
--- a/platform/dummy_platform.go
+++ b/platform/dummy_platform.go
@@ -184,7 +184,7 @@ func (p dummyPlatform) SetupRawEphemeralDisks(devices []boshsettings.DiskSetting
 	return
 }
 
-func (p dummyPlatform) SetupDataDir() error {
+func (p dummyPlatform) SetupDataDir(_ boshsettings.JobDir) error {
 	dataDir := p.dirProvider.DataDir()
 
 	sysDataDir := filepath.Join(dataDir, "sys")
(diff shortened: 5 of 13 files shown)


voelzmo:
This looks nice! Just to help me understand the motivation: Is this to avoid writing secrets to disk which are needed for configuring a job? Do you have other motivations for this change?

johnpfield:
That's the idea.  The requirement we're trying to address is to avoid persisting sensitive data (like, say, a certificate's private key, to a disk).  The credential would exist only in memory.  This will enable deployers who have an InfoSec policy against writing credentials in the clear to disk, to be compliant with that policy.

voelzmo:
Thanks @johnpfield for clarifying! And the attack vector that you're addressing here is an untrusted provider of storage? If you have access to the machine itself, it doesn't make a difference, correct?

johnpfield:
Agreed.  That's a reasonable characterization.  A privileged user can still get the values when on the host.  But persisting to disk implies that we need to trust that the disk itself will always be safe from misuse.  If credentials lived only in tmpfs/ramfs then they will never persist beyond the associated process, in time or space.  If written to disk, I might later mount that (maybe orphaned) disk on another VM, for example.

voelzmo:
Thanks for the explanations, all of this makes a lot of sense to me!

voelzmo:
@xoebus one more thing about the tmpfs: How do you imagine dealing with reboots? Should the agent get new templates every time it starts?

xoebus:
We decided to let health-monitor do it's thing to begin with and work out
how to handle rebooted machines later.

dpb587-pivotal:
*caveat that I'm not sure health monitor will currently recreate in the case where process state is empty. I think it currently only recreates if the VM is entirely missing. It might require a further change in agent or health monitor for testing this.

jfmyers9:
Hi @xoebus,

We took a look at this, and from a code perspective it seems fine.

Our only concern is what happens during a reboot. As @dpb587-pivotal pointed out, the health monitor only recreates VMs when the VM is actually missing. Therefore reboots still will not be successful. This is definitely an issue for infrastructures in which reboots are more common (i.e. `vSphere`).

I'm hesitant to update the health monitor to recognize this scenario as it could cause unnecessary / unwanted recreates. For example, an empty instance group will have no `/var/vcap/jobs` directory at all.

It feels like a real solution would be for the agent to ping the director and have the director reinitialize the VM if necessary. This seems much more complicated as we would now need to concern ourselves with creating a communication channel from the agent -> director and securing it.

Another solution would be to have the director recognize reboots, and handle them. I'm not sure what this would actually involve at the moment.

Thoughts??

@jfmyers9 && @mikexuu, CF BOSH

xoebus:
There could be benefit to removing the restart case from the BOSH instance state machine altogether. If an instance restarts then the agent can refuse to start up and force the instance to be recreated. I don't think this would require changes in the health monitor.

This would also fix the `prestart` gotchas with some jobs not expecting a state change directly from _not-running_ to _running_ directly without passing through the _prestart_ state which occurs when a machine is restarted.

What do you think?

jfmyers9:
How would an agent detect that it's VM has been restarted?

Also I'm worried that by refusing to start the agent on reboot, that could result in expensive and undesired behavior. Restarting the processes on the VMs that have been rebooted would now be blocked behind a recreate of every VM in question which is costly and time consuming. In the case in which a large number of VMs were rebooted (i.e. system maintenance?), the time to recovery could be potentially much larger.

I don't have any actual data on how common reboots are in production environments, but the VM thrash of having to recreate virtual machines on reboot might also be undesired.

xoebus:
> How would an agent detect that it's VM has been restarted?

It could write a sentinel on first startup and then check if that exist on subsequent starts. The method isn't important though.

> I don't have any actual data on how common reboots are in production environments, but the VM thrash of having to recreate virtual machines on reboot might also be undesired.

Let's ask the field about this.

Having the agent ping the director worries me because then potentially anyone can ping the director and be delivered some new credentials. We could handle this by doing the machine identity work. That seems pretty far out at the moment though.

xoebus:
How about I add another feature flag for this and we can try it out?

jfmyers9:
Where would we reasonably be able to try this out? I don't think developers are the ones that often run into the VM reboot situation, this generally occurs at the end user / production environments. I'd almost rather us just implement the `tmpfs` for the time being and have this be a caveat for users of that experimental feature.

Rather than having the agent reach out to the director, we could explore improving the granularity of what the health monitor notices. Can the health monitor detect that a VM has been rebooted and then instruct the director to "re-prepare" the VM? Seems like a more comprehensive solution.

xoebus:
> Rather than having the agent reach out to the director, we could explore improving the granularity of what the health monitor notices. Can the health monitor detect that a VM has been rebooted and then instruct the director to "re-prepare" the VM? Seems like a more comprehensive solution.

Does the agent have a ping or announce message which it sends when it launches? We could include the uptime of the machine in that ping which would let us detect if the agent identifier stays the same but the uptime decreases. However, NATS doesn't guarantee in-order delivery so we don't want to reboot the machine if two ping messages appear out of order. We'll need to include a timestamp too.

However, this introduces a whole new health-monitor concept of trying to repair machines rather than just blasting them away. This is against the BOSH ethos of being ruthless with machines. It also introduces a whole new control path which will rarely be exercised. This means it'll be expensive to test and maintain. Does this also add a new failure mode which jobs will need to handle?

I'd still much rather simplify and unify things and then improve the experience of using those few paths. I'm not sure if the experience of recreating the machine would be any worse than having the health-monitor perform some pseudo soft-recreate.

xoebus:
In any case, I'll tidy this PR up into a complete change.

dpb587-pivotal:
I'd prefer we avoid changing the (director|health monitor)<=>agent API contracts to test this out. If we can rely on existing data/methods, I think that would help keep the scope of this small.

I'd strongly prefer to avoid agent being able to request re-provisioning of itself. There are security implications of that, but it also diverges from BOSH's model of director being the one to initiate all aspects of provisioning. Even if introducing a "reimage" task which relies on an instance already existing, I'd prefer relying on recreate behavior to ensure more of a guaranteed start position.

When a VM restarts, I could imagine some IaaS environments cause you to lose or change state of things. Network configuration comes to mind, but I don't have specific scenarios. In general though, can we absolutely say and assume that, after a restart, on every IaaS, the state of things on the VM and its network+disk connectivity were exactly as they were created? If not, it would be safer to recreate the VM from scratch.

Recreating from scratch when a VM goes down would be expensive in the world where an entire rack goes down. It currently is extremely inefficient, especially since resurrection requires a deployment lock and only recovers one VM at a time. That constraint would still be in effect for the "reinitialize me" agent ping, less the IaaS resource management time (which is notable).

I see value in supporting an "disabled restart" behavior in the agent (initially implied by this `tmpfs` flag, but could theoretically be exposed separately, later). It would be a much smaller and independent change to make agent `halt`/`sv stop`/whatever if it detects a restart, and then we can rely on current health monitor behavior (it already has a shutdown manager somewhere for the director/agent `shutdown` API call). This behavior would certainly not be long-term optimal, but I think its simplicity would still be sufficient for an MVP of this approach. I think we should at least try out this approach.

johnpfield:
@jbayer ICYMI, this was the pull request done by @xoebus

PR #188, opened 2018-08-29
Make integration tests 11x faster (44m -> 4m)

xoebus:
This patch opens an SSH connection once at the start of the test suite rather than invoking `vagrant ssh` thousands of times. Each `vagrant ssh` has around one second of overhead which built up. The tests are 11x faster on my machine with this patch applied.

I had to change one test because it was expecting to be run in a shell environment with pipes. That was the only test which depended on that behavior.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -150,6 +150,12 @@
   version = "v1.1.0"
 
 [[projects]]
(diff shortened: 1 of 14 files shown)


dpb587-pivotal:
I needed to run agent integrations last week and re-appreciated this pull request. Thanks, again :heart:

PR #195, opened 2018-10-24
Simplify NicSettingsTemplate

benmoss:
We were running into a problem where after creating an HNS Network we were getting two network adapters returned for the same MAC address.

This was then causing the `netsh interface ip set` line of the NicSettingsTemplate powershell script to fail since it was combining the two names.

We tried this change out with a 2019 stemcell. It fixed our problem and seemed to otherwise work fine.

@ashwin-venkatesh

### DIFF
--- a/platform/net/windows_net_manager.go
+++ b/platform/net/windows_net_manager.go
@@ -86,8 +86,7 @@ foreach($interface in $interfaces) {
 `
 
 	NicSettingsTemplate = `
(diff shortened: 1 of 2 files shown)


ashwin-venkatesh:
Hey @aminjam and @sesmith177 
Any idea if this change impact y'all in any way? This is some super old code and I'm OK with merging it if nothing seems obviously wrong

aminjam:
Looks good! I don't see anything that breaks us. I don't know why previously it was running `Get-WmiObject` if we have the Name available all along.

PR #196, opened 2018-10-25
Make fly exec to not ignore bosh-agent binary

beyhan:
Currently, when running the repack-stemcell script the concourse one-off ignores the just created bosh-agent binary. 

We temporarily added the bosh-agent binary to the git index to ensure it is being picked by concourse as input resource. Alternatively the concourse CLI provides a flag `fly execute --include-ignored` which will take everything ignored by git. However, we thought there might be credentials you do not want to be uploaded to concourse.

Any thoughts from your side?
 
Co-authored-by: Florian Nachtigall <florian.nachtigall@sap.com>

### DIFF
--- a/bin/repack-stemcell/run.sh
+++ b/bin/repack-stemcell/run.sh
@@ -10,6 +10,10 @@ cd "${bin}"/../..
 
 mv out/bosh-agent bin/bosh-agent # necessary so that fly -x can be used
 
+git add -f bin/bosh-agent
+
 time fly -t production execute -p -i agent-src=. -o stemcell=out/ -c ./bin/repack-stemcell/task.yml
+
+git reset bin/bosh-agent
 
 ls -la out/stemcell.tgz


jfmyers9:
Hi @beyhan && @FlorianNachtigall,

My biggest concern with the proposed implementation is that if the `fly execute` fails the `git reset` will not run. This means that there could be a situation in which the `bosh-agent` binary remains a tracked file. I would worry that this would make it easier for users to check in the binary.

I would prefer using the `--include-ignored` flag at the moment. The concourse that we are generally targeting is secure and usually has access to the credentials that we may have lying around (pipeline credentials?). Are there any other concerns that you have with that implementation? This would also allow us to add `bin/bosh-agent` to the `.gitignore` so that we prevent users from ever checking in a binary to the repository.

Thoughts?

Best,

@jfmyers9 && @langered, CF BOSH

dpb587-pivotal:
-1 from me as well due to same concerns.

Other options are either introducing a new input which doesn't worry about its `.gitignore` (the script could create its own tmpdir with the file, perhaps), or just using a different, non-ignored path.

Closing, but happy to continue discussing ideas.

PR #177, opened 2018-07-02
DNS servers for preconfigured networks

edwardstudy:
This PR worked on configuring DNS servers for preconfigured networks.

We merged PR 151: https://github.com/cloudfoundry/bosh-agent/pull/151#issuecomment-381711516 and modified this code https://github.com/cloudfoundry/bosh-agent/blob/c74db775183f3e0e68254279e649b802bd4bf63f/platform/net/ubuntu_net_manager.go#L100-L103
It skipped configuring network alias interfaces and caused empty virtual interfaces. We should continue to configure networked after writing DNS entries. 

@mikexuu Please help to view about network configs.

PR #199, opened 2018-11-27
Remove writeresolvConf func

edwardstudy:
According to [**comment**](https://github.com/cloudfoundry/bosh-agent/pull/177#issuecomment-440699644), bosh-agent doesn't need to write `resolv.conf` that is duplicate action because agent finally write DNS records on interfaces.

Be careful with bosh-lite which VMs will haven't DNS records because they are defined `preconfigured`. 

Best,

@edwardstudy, IBM

/cc @mattcui @dpb587-pivotal


edwardstudy:
Yes, I also cared about other IaaSes settings.

As I said, bosh-lite VM didn't get right DNS settings if we removed `writeresolvConf` func. 

We can talk about this next sync meeting.

edwardstudy:
Still reserved writeresolvConf func for `preconfigured`.

And ned to reduce operations for symbolic links under writeresolvConf func.

edwardstudy:
Hi, @dpb587-pivotal. I had to raise another [PR](https://github.com/cloudfoundry/bosh-agent/pull/202) to reduce duplicated code.

PR #200, opened 2018-12-05
162161084 fix windows symlink

xtreme-jason-smith:


xtreme-jason-smith:
Note: This will break the BOSH Agent as it requires BOSH Utils changes. There is a pull request to update the BOSH Utils with this PR: https://github.com/cloudfoundry/bosh-utils/pull/52

agent/applier/bundlecollection/file_bundle_windows_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"os"
+
+	. "github.com/cloudfoundry/bosh-agent/agent/applier/bundlecollection"
+	"github.com/cloudfoundry/bosh-agent/agent/applier/bundlecollection/fakes"
+	fakefileutil "github.com/cloudfoundry/bosh-utils/fileutil/fakes"
+	boshlog "github.com/cloudfoundry/bosh-utils/logger"
+	fakesys "github.com/cloudfoundry/bosh-utils/system/fakes"
+)
+
+//go:generate counterfeiter -o fakes/fake_clock.go ../../../vendor/code.cloudfoundry.org/clock Clock
```
dpb587-pivotal: Is this something different than https://github.com/cloudfoundry/bosh-agent/tree/master/vendor/code.cloudfoundry.org/clock/fakeclock?

xtreme-jason-smith: Nope, it should be the same one. We copied part of the code from the file_bundle_test.go

xtreme-jason-smith:
I forgot to mention that this commit does not include the go dep changes to pull in the BOSH Util code. Will that be taken care of automatically or should I update the pull request to include the dep changes?

PR #202, opened 2018-12-13
Reduce duplicated code for writeresolvConf func

edwardstudy:
`writeresolvConf` func had duplicated code about `etc/resolv.conf` symbolic link which `resolvcof`  package already covered. This PR reduced duplicated code.

Best,

@edwardstudy, IBM

/cc @mattcui @dpb587-pivotal

### DIFF
--- a/platform/net/ubuntu_net_manager.go
+++ b/platform/net/ubuntu_net_manager.go
@@ -3,7 +3,6 @@ package net
 import (
 	"bytes"
 	"path"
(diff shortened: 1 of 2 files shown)


luan:
@edwardstudy friendly nudge :)

edwardstudy:
@luan Hi, Sorry for late. This PR is a low priority and only targets to reduce duplicated commands.

PR #203, opened 2018-12-17
Add jitter to logrotate cron schedule

jaresty:


pivotal-mp:
This was merged already on another branch.

PR #212, opened 2019-03-11
Use `env` options in `runScript` 

charleshansen:
to set environment variables in script runtime

- To support `pre-stop` hook
[#164262378](https://www.pivotaltracker.com/story/show/164262378)

Co-authored-by: Belinda Liu <bliu@pivotal.io>
Co-authored-by: Charles Hansen <chansen@pivotal.io>

Note that there a lots of files changing because the linter ran. 

The important parts are in `run_scripts.go`, `generic_script.go` and their tests. The `concrete_job_script_provider.go` also had to change a little because of the interface change.

There are currently no integration tests since that would require some new architecture that is only available for the windows tests.


aashah:
Yeah looks like the integration tests are a bit messy. I think we'll take some time trying to untangle them for next PRs :)

PR #213, opened 2019-03-13
Use a defined type rather for script options

jfmyers9:
This change updates `run_script` to take use a structured object for
options rather than an genreic map. This change should be both backwards
compatible and forwards compatible due to the unmarshalling behavior of
JSON in Golang. Adding fields to the options struct will be additive
changes and will not break older directors.

[#164262374](https://www.pivotaltracker.com/story/show/164262374)


dpb587-pivotal:
Good idea.

PR #215, opened 2019-03-26
Prevent logging of reading records.json

NautiluX:
* whenever agent received sync_dns action, it logged the content of the
records.json.
* In the case where an update of the dns records is needed on the VM, the file content is logged 3 times.
* With a large number of VMs managed by BOSH this led to frequent log rotations
making logs unusable. (With more than 1,000 VMs, agent logs got
lost after 15 minutes)

[#164371973](https://www.pivotaltracker.com/story/show/164371973)

Co-authored-by: David Ansari <david.ansari@sap.com>

### DIFF
--- a/agent/action/state/sync_dns_state.go
+++ b/agent/action/state/sync_dns_state.go
@@ -65,7 +65,7 @@ func (s SyncDNSState) NeedsUpdate(newVersion uint64) bool {
 }
 
 func (s SyncDNSState) loadVersion() (uint64, error) {
-	contents, err := s.fs.ReadFile(s.path)
+	contents, err := s.fs.ReadFileWithOpts(s.path, boshsys.ReadOpts{Quiet: true})
 	if err != nil {
 		return 0, bosherr.WrapError(err, "reading state file")
 	}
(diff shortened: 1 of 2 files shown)


aashah:
Hey,

Spoke with folks on the team. They agree that is 100% way too spammy, but that we've also found useful during debugging/troubleshooting.

It looks like the contents of `records.json` won't be emitted after this PR at all. We're wondering if it's still useful to print it once, maybe on updates?

Otherwise we are totally happy merging this.

- Aakash & @SocalNick

NautiluX:
Hi @aashah, @SocalNick,

thanks for looking at the PR. We adjusted it according to your proposal.

Independent of this specific log, we wondered if there are any plans on making the minimal logged level of the agent configurable? Currently it is hardcoded [here](https://github.com/cloudfoundry/bosh-agent/blob/59ca8116eae8e76af9898ca970b765a44bbc7919/main/agent.go#L78) to `Debug`.

Manuel & @ansd

NautiluX:
Thanks for merging it in. I will close this PR then.

PR #217, opened 2019-04-19
Enhancement for encryption of persistent disk

zts-pash:
This pull request includes the changes required to support usage and management of the encrypted disk in the bosh environment. Since the mount has to be done differently when the persistent disk is encrypted which needs encryption keys as well, so therefore it has to be handled externally (during deployment of encryption job) and not by the bosh-agent.
Seeing issues for scenarios like redeployment of vms or deletion/creation of vm having the same disk volume where bosh was trying to mount the already encrypted disk which now has fs type as crypto_LUKS instead of ext4/xfs.

Currently in bosh-agent code there is a method which returns cid to bosh director if the disk is mounted correctly. Since bosh is not able to mount the encrypted disk correctly in the above mentioned scenrio, therefore it is returning nil to the director, where AgentDiskOutOfSync check  compares the cid stored within the director to the one received by the agent which are now different in this case and therfore it raises exception saying "instance has invalid disks: agent reports 'nil' while director record shows vol-xxxx".

To fix this issue, the code change is done entirely on the bosh-agent side where it checks if the partition format type is crypto_LUKS through lsblk command (as suggested last in previous pull request), then don't try to mount the partition which will be done by the job later on and the variable IsMountedExternally is set to true in this case.
One more filesystem type is added i.e crypto_LUKS apart from ext4 and xfs and now even at the time of formatting the disk this new fs type is also checked along with other fs types. If the disk is already formatted with crypto_LUKS type, it shouldn't be formatted again.

We tried to make this code change generic enough by implementing it through interface level in such a way that anyone who wants to handle mounting of partition externally can leverage this implementation.

Co-authored-by: Maksim Yankovskiy <myankovskiy@zettaset.com>
Co-authored-by: Deepak Shetty <dshetty@zettaset.com>
Co-authored-by: Brian Metzger <bmetzger@zettaset.com>

### DIFF
--- a/agent/action/list_disk.go
+++ b/agent/action/list_disk.go
@@ -61,11 +61,11 @@ func (a ListDiskAction) Run() (interface{}, error) {
 		if err != nil {
 			return nil, bosherr.WrapErrorf(err, "Checking whether device %+v is mounted", diskSettings)
 		}
-                isMountedExternally, err := a.platform.IsPersistentDiskMountedExternally(diskSettings)
-                if err != nil {
+		isMountedExternally, err := a.platform.IsPersistentDiskMountedExternally(diskSettings)
+		if err != nil {
 			return nil, bosherr.WrapErrorf(err, "Checking whether device %+v is mounted externally", diskSettings)
 		}
-		if isMounted || (!isMounted && isMountedExternally) {
+		if (isMounted || isMountedExternally) {
 			if _, present := usedIDs[diskID]; !present {
 				diskIDs = append(diskIDs, diskID)
 				usedIDs[diskID] = true
--- a/agent/action/list_disk.go
+++ b/agent/action/list_disk.go
@@ -61,8 +61,11 @@ func (a ListDiskAction) Run() (interface{}, error) {
 		if err != nil {
 			return nil, bosherr.WrapErrorf(err, "Checking whether device %+v is mounted", diskSettings)
 		}
-
-		if isMounted {
+                isMountedExternally, err := a.platform.IsPersistentDiskMountedExternally(diskSettings)
+                if err != nil {
+			return nil, bosherr.WrapErrorf(err, "Checking whether device %+v is mounted externally", diskSettings)
+		}
+		if isMounted || (!isMounted && isMountedExternally) {
 			if _, present := usedIDs[diskID]; !present {
 				diskIDs = append(diskIDs, diskID)
 				usedIDs[diskID] = true
--- a/platform/disk/linux_mounter.go
+++ b/platform/disk/linux_mounter.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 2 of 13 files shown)


platform/disk/linux_mounter.go
```diff
@@ -147,6 +148,15 @@ func (m linuxMounter) IsMountPoint(path string) (string, bool, error) {
 	return "", false, nil
 }
 
+func (m linuxMounter) IsCryptLuks(partitionOrMountPoint string) (bool, error) {
+	out, _ := exec.Command("sh","-c","lsblk -f | grep crypto_LUKS | wc -l").Output()
```
aashah: Handling the error would be nice here.

platform/disk/parted_partitioner.go
```diff
@@ -157,8 +158,11 @@ func (p partedPartitioner) GetPartitions(devicePath string) (partitions []Existi
 			return partitions, deviceFullSizeInBytes, bosherr.WrapErrorf(err, "Parsing existing partitions")
 		}
 
+		out, _ := exec.Command("sh","-c","lsblk -f | grep crypto_LUKS | wc -l").Output()
```
aashah: Handling the error would be nice here.

platform/linux_platform.go
```diff
@@ -1169,12 +1169,17 @@ func (p linux) MountPersistentDisk(diskSetting boshsettings.DiskSettings, mountP
 		realPath = partitionPath
 	}
 
-	err = p.diskManager.GetMounter().Mount(realPath, mountPoint, diskSetting.MountOptions...)
-
+	existingFsType, err := p.getPartitionFormatType(partitionPath)
```
aashah: Instead of duplicating the code from `linuxFormatter#getPartitionFormatType`, could we use the `#IsPersistentDiskMountedExternally` introduced in this same file? Seems more robust for future externally mounted disks.

agent/action/list_disk.go
```diff
@@ -61,8 +61,11 @@ func (a ListDiskAction) Run() (interface{}, error) {
 		if err != nil {
 			return nil, bosherr.WrapErrorf(err, "Checking whether device %+v is mounted", diskSettings)
 		}
-
-		if isMounted {
+                isMountedExternally, err := a.platform.IsPersistentDiskMountedExternally(diskSettings)
+                if err != nil {
+			return nil, bosherr.WrapErrorf(err, "Checking whether device %+v is mounted externally", diskSettings)
+		}
+		if isMounted || (!isMounted && isMountedExternally) {
```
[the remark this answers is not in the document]
zts-pash: Thanks Aakash for your review and feedback. Will work on it soon.

platform/linux_platform.go
zts-pash: I pushed the changes you suggested. Also the answers to your other question is yes. disk will be mounted by the co-located (luks addon) job.

PR #222, opened 2019-09-27
Need to wipe off both partitions and devices

mattcui:
The original PR is https://github.com/cloudfoundry/bosh-agent/pull/190, but we found after OS reload, the ephemeral disk is not re-formatted, because it only wipes off the device, but doesn't wipe off partitions. This PR is very important to our IBM Cloud, the disk is not cleaned up  after stemcell upgrade, which caused many problems in our cf deployments. Please help review this PR. Thanks.

/cc @dpb587 @maximilien


mattcui:
Did anyone review the PR? Thanks.

PR #225, opened 2019-11-20
fix virtual interface systemd network file

gu-bin:


aashah:
Hey @gu-bin,

Like mentioned in the issue, we approached the solution from a different angle.

Going to close this PR for now.

Thanks

PR #229, opened 2020-03-04
Fix vitals output when no persistent disk present

aashah:
https://github.com/cloudfoundry/bosh/issues/2204

### DIFF
--- a/platform/provider.go
+++ b/platform/provider.go
@@ -71,7 +71,7 @@ func NewProvider(logger boshlog.Logger, dirProvider boshdirs.Provider, statsColl
 	// Kick of stats collection as soon as possible
 	statsCollector.StartCollecting(SigarStatsCollectionInterval, nil)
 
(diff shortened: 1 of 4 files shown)


platform/provider.go
```diff
@@ -143,7 +143,7 @@ func NewProvider(logger boshlog.Logger, dirProvider boshdirs.Provider, statsColl
 			compressor,
 			copier,
 			dirProvider,
-			vitalsService,
+			linxuVitalsService,
```
aashah: I think you may need to pull your bosh-agent code. Centos has been removed as a build target for the bosh-agent.

platform/provider.go
```diff
@@ -221,7 +221,7 @@ func NewProvider(logger boshlog.Logger, dirProvider boshdirs.Provider, statsColl
...
```
aashah: Same here, no Suse.

Issue #230, opened 2020-03-26
Bosh agent paniced during host routing parsing


hanlins:
# Environment
Stemcell: `bosh-azure-hyperv-ubuntu-xenial-go_agent/621.59`
Opsman: `v2.8.5-build.234`

# Description
Deploying `calico` in a kubernetes cluster as pod network solution. Calico will deploy `calico-node` on worker VMs as a daemon set, which will modify the VM network setting (adding tunneling interface, add routes). As the VM host routing got modified, bosh agent failed parsing host routing, and emit following panic logs in agent log (at `/var/vcap/bosh/log/current`):
```
22:05:43 DEBUG - Successful: true (0)
2020-03-26_22:05:43.90478 panic: runtime error: index out of range [0] with length 0
2020-03-26_22:05:43.90481
2020-03-26_22:05:43.90481 goroutine 18 [running]:
2020-03-26_22:05:43.90481 github.com/cloudfoundry/bosh-agent/platform/net.parseRoute(0xc0002b034c, 0x28, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
2020-03-26_22:05:43.90482 	/tmp/build/9674af12/gopath/src/github.com/cloudfoundry/bosh-agent/platform/net/routes_searcher_unix.go:29 +0x425
2020-03-26_22:05:43.90482 github.com/cloudfoundry/bosh-agent/platform/net.cmdRoutesSearcher.SearchRoutes(0xbd3680, 0xc000166770, 0xaa7760, 0xc0001a68c0, 0xc0001a6958, 0x989833, 0x2)
2020-03-26_22:05:43.90482 	/tmp/build/9674af12/gopath/src/github.com/cloudfoundry/bosh-agent/platform/net/routes_searcher_unix.go:60 +0x19e
2020-03-26_22:05:43.90482 github.com/cloudfoundry/bosh-agent/platform/net.defaultNetworkResolver.GetDefaultNetwork(0xbbf9c0, 0xc000166820, 0xbbfa80, 0xafba58, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
2020-03-26_22:05:43.90484 	/tmp/build/9674af12/gopath/src/github.com/cloudfoundry/bosh-agent/platform/net/default_network_resolver.go:29 +0xa5
[...]

PR #231, opened 2020-03-27
Fix blackhole route parsing

hanlins:
In this fix, perform checking before consuming the route parsing results to avoid [panic issue](https://github.com/cloudfoundry/bosh-agent/issues/230).

### DIFF
--- a/platform/net/routes_searcher_unix.go
+++ b/platform/net/routes_searcher_unix.go
@@ -67,7 +67,7 @@ func (s cmdRoutesSearcher) SearchRoutes() ([]Route, error) {
 		}
 		route, err := parseRoute(routeEntry)
 		if err != nil {
-			s.logger.Warn("SearchRoutes", "parseRoute error: %s", err.Error())
+			s.logger.Warn("SearchRoutes", "parseRoute error for route '%s': %s", routeEntry, err.Error())
 			continue
 		}
 		routes = append(routes, route)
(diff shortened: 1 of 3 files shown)


platform/net/routes_searcher_unix.go
```diff
@@ -20,10 +20,14 @@ func NewRoutesSearcher(runner boshsys.CmdRunner, _ InterfaceManager) RoutesSearc
 	return cmdRoutesSearcher{runner}
 }
 
-func parseRoute(ipString string) Route {
+func parseRoute(ipString string) (Route, error) {
 	var r = regexp.MustCompile(`(?P<destination>[a-z0-9.]+)(/[0-9]+)?( via (?P<gateway>[0-9.]+))? dev (?P<interfaceName>[a-z0-9]+)`)
 
 	match := r.FindStringSubmatch(ipString)
+	// Issue #230, skip blackhole routes
```
h4xnoodle, line 27: could you please remove the comment? the issue will already be tied via the PR/commit.

platform/net/routes_searcher_unix.go
```diff
@@ -57,7 +61,11 @@ func (s cmdRoutesSearcher) SearchRoutes() ([]Route, error) {
 		if len(routeEntry) == 0 {
 			continue
 		}
-		routes = append(routes, parseRoute(routeEntry))
+		route, err := parseRoute(routeEntry)
+		if err != nil {
+			continue
```
h4xnoodle, line 66: could we log that this happened? in case it's unexpected that we couldn't parse the route

h4xnoodle:
closing and reopening to hopefully get rid of a bot

PR #232, opened 2020-04-08
Revert "Fix blackhole route parsing"

aashah:
Reverts cloudfoundry/bosh-agent#231

Original PR was targeting wrong branch.

### DIFF
--- a/platform/net/routes_searcher_unix.go
+++ b/platform/net/routes_searcher_unix.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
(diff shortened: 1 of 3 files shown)


aashah:
@hanlins 

I am gonna have to revert the origin PR since it was targeting master.

Also, test-unit was failing due to compilation errors, which didn't get caught because our CI doesn't run against the master branch.

hanlins:
@aashah which branch should it be targeted at? We need the fix ASAP.

mrosecrance:
Aakash did merge it onto develop. We're having some pipeline issues to sort out but it will be automatically bumped to master once it passes.

PR #234, opened 2020-06-25
Increase size of /var/vcap/data/sys/run and make configurable

krutten:
[172950654](https://www.pivotaltracker.com/story/show/172950654)

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -130,6 +130,10 @@ func (boot bootstrap) Run() (err error) {
 		return bosherr.WrapError(err, "Starting up logging and auditing utilities")
 	}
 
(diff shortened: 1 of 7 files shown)


mrosecrance:
@krutten This change will need unit and integration tests before being merged. Also this line: https://github.com/cloudfoundry/bosh-agent/blob/1710e5c5a0ccf606b3af4d75ef2c41820641af81/platform/linux_platform.go#L828 looks like it's also being run on /sys/run/ and seems like it's now redundant and perhaps we can remove the method.

rkoster:
closing in favour of https://github.com/cloudfoundry/bosh-agent/pull/236

PR #238, opened 2020-12-11
workaround when netplan is a hard requirement for the IAAS

gu-bin:
For Softlayer, the Bionic stemcell image must have `netplan` installed, otherwise the VM can't be provisioned.

Softlayer uses `netplan` to config the network of Bionic VMs which generates the following two .network files for `systemd-networkd`
```
/run/systemd/network/10-netplan-eth0.network
/run/systemd/network/10-netplan-eth1.network
```
These two files are processed in higher priority than the two generated by `bosh-agent`
```
/etc/systemd/network/10_eth0.network
/etc/systemd/network/10_eth1.network
```
So if a virtual interface like `eth0:x` is used which is only configured in the bosh-agent generated files, it can't take effect.

This PR changed the prefix of .network files generated by `bosh-agent` from `10_` to `09_`, so they can be processed in higher priority than the ones generated by `netplan`. 

One situation needs to be avoid. Before the VM network configuration is ready, `bosh-agent` can't get the network information from VM settings, so it will generate a `09_eth0.network` file with empty settings. As this file takes higher priority, when `systems-networkd` is restarted, the VM will have empty network configuration. This will make the VM provision failed. So this PR also make `09_eth0.network` not to be generated until network configuration is ready.

This Slack discussion is here https://cloudfoundry.slack.com/archives/C01C7V4U79U/p1607317868022400


mrosecrance:
Could we use a higher sub 10 number - like 9? It may never be used but I'd like to keep the option open in case there is ever a subcase in which we'd need to overwrite the priority.

mrosecrance:
Also you should switch the pr branch to develop (master is automatically changed when tests pass)

gu-bin:
@mrosecrance I've addressed your comments and updated the PR description as well. Please review. Thanks.

ramonskie:
we just ran in to issues regarding your PR
with aws,gcp,alicloud
as the network is never created.
it seems to me that netplan is specifically installed in softlayer here https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-bionic/master/stemcell_builder/stages/system_softlayer_netplan/apply.sh
if your problem is netplan.
than we should revert your pr on the bosh-agent.
and remove the stage system_softlayer_netplan
in the bosh-stemcell-builder.

PR #239, opened 2021-01-22
Fix double-resize persistent disk on vSphere

mariash:
When extending the size of a persistent disk _for the second time_ on vSphere, the operation fails with "...Creating partition using parted...". This commit addresses that error.

Our fix is to strategically run `partprobe`, which "informs the operating system kernel of partition table changes".

We believe the cause of this error is the manner in which vSphere detaches disks from VMs, which differ from the manner in which other IaaSes, specifically GCP, detach disks. vSphere does not signal to the OS that the disk has been removed, and the memory of the disk "lingers".

For example, running `lsblk -a` after the first resize (say from 1GB → 2GB), the new (larger) disk shows up as `/dev/sdd` and the old (smaller) disk shows up as `/dev/sdc`. This stands in sharp contrast to GCP, where the new (larger)
disk shows up as `/dev/sdc`, and the old (smaller) disk is not to be found.

Note that after the first resize, all is good on vSphere: `/dev/sdd1` is mounted on `/var/vcap/store` (on GCP it's `/dev/sdc1` that's mounted on `/var/vcap/store`). It's when the second resize (say from 2GB → 3GB) is done that everything breaks down on vSphere: the new 3GB disk is connected as `/dev/sdc`, but as far as the kernel is concerned `/dev/sdc` is not the new 3GB disk but the original 1GB disk with the original partition table, and the `bosh-agent` errors saying the disk is too small.

Fixes, on `bosh deploy`:
```
Task 28327 | 20:05:31 | L executing post-stop: app/cad4e5cd-cf79-4af1-9d50-2393109227a8 (0) (canary) (00:03:29)
                     L Error: Action Failed get_task: Task 36a76be2-0c67-4c07-78d7-bb6d44e71344 result: Mounting persistent disk: Partitioning disk: Partitioning disk `/dev/sdc': Creating partition using parted: Running command: 'parted -s /dev/sdc unit B mkpart bosh-partition-0 1048576 2146435071', stdout: '', stderr: 'Warning: Not all of the space available to /dev/sdc appears to be used, you can fix the GPT to use all of the space (an extra 2097152 blocks) or continue with the current setting?
Error: You requested a partition from 1048576B to 2146434560B (sectors 2048..4192255).
The closest location we can manage is 1048576B to 1073724416B (sectors 2048..2097118).
```

### Please Note

- We're looking for guidance about which branch to submit this PR to. We're submitting against develop, but we suspect it should be backported to branch `2.268.x`
- We tested this fix by applying it as a patch to the `2.268.16` tag and deploying three persistent disk resizes in a row.
- We ran the unit tests, but were unable to run the integrations tests.

/cc @jpalermo who did much of the initial investigation

### DIFF
--- a/platform/disk/ephemeral_device_partitioner_test.go
+++ b/platform/disk/ephemeral_device_partitioner_test.go
@@ -84,7 +84,7 @@ var _ = Describe("EphemeralDevicePartitioner", func() {
 				err := partitioner.Partition(devicePath, partitions)
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(len(fakeCmdRunner.RunCommands)).To(Equal(12))
+				Expect(len(fakeCmdRunner.RunCommands)).To(Equal(13))
 
 				Expect(fakeCmdRunner.RunCommands).To(ContainElement([]string{"parted", "-m", "/dev/edx", "unit", "B", "print"}))
 				Expect(fakeCmdRunner.RunCommands).To(ContainElement([]string{"parted", "-s", "/dev/edx", "unit", "B", "mkpart", "fake-agent-id-0", "1048576", "8590983167"}))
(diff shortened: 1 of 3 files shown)


mrosecrance:
@cunnie Mind signing the CLA? 

2.268.x is the agent version of 621 so you're correct that that would be the line to backport. Did you all have a conversation about getting vsphere to signal the OS or is that going to be really tricky? The code change does seem really small, but it's probably going to take us a while to get to a deeper look at this PR.

cunnie:
> @cunnie Mind signing the CLA?

Done!

> 2.268.x is the agent version of 621 so you're correct that that would be the line to backport.

@mrosecrance **Should we submit a second PR for that line?**

> **Did you all have a conversation about getting vsphere to signal the OS or is that going to be really tricky?** The code change does seem really small, but it's probably going to take us a while to get to a deeper look at this PR.

That's going to be tricky—we wouldn't even know which team to ask. Also, I'm not 100% sure that this is a vSphere "bug". For example, when I disconnect a SCSI device from a SCSI bus (which, IIRC, vSphere is emulating), there's no signal sent to the OS that a device has been detached—the device merely silently disappears. Which is the behavior that vSphere is emulating.

PR #246, opened 2021-09-08
Have vitals check for mount points (fixes #245)

### LINKED ISSUE #245: bosh-agent reporting erroneous disk data on bionic
BOSH Director: `271.9.0`
CPI: `AWS/88`
Stemcell: Bionic `1.22`

We moved various deployments off stemcell Xenial `621.125` to Bionic `1.22` and our Prometheus system started to alert on persistent disk usage warnings on various VMs which do not have any persistent disk assigned.  Our alert threshold was set to 75% and most alerts came in between 75.1 and 75.9 after moving the VMs to Bionic.

Looking at the VMs in question I found the following under `/var/vcap/bosh`

`spec.json`
```
  "persistent_disk": 0,
```

`settings.json`
```
    "persistent_disk_fs": "",
    "persistent_disk_mount_options": null,
    "persistent_disk_partitioner": ""
```
`/var/vcap/bosh/log/current`
```
2021-09-01_13:41:22.72666 {"deployment":"cf-production","job":"nats","index":2,"job_state":"running","vitals":{"cpu":{"sys":"0.7","user":"1.7","wait":"0.1"},"disk":{"ephemeral":{"inode_percent":"6","percent":"19"},"persistent":{"inode_percent":"36","percent":"76"},"system":{"inode_percent":"36","percent":"76"}},"load":["0.00","0.06","0.07"],"mem":{"kb":"2426640","percent":"30"},"swap":{"kb":"0","percent":"0"},"uptime":{"secs":87633}},"node_id":"XXXXXXXX"}
```

Since we changed stemcell these where fresh VMs but a `bosh recreate` did not resolve the issue.  We tried to add additional ephemeral disk as well, no change in behavior.  The only way to make the stat drop was to add a persistent disk to the VMs in question.

This appears to have happened before with the `bosh-agent` but these issues suggested a possible code fix and/or nothing done and the issues auto closed for being stale:

https://github.com/cloudfoundry/bosh/issues/2204

https://github.com/cloudfoundry/bosh-agent/issues/36

I'm wondering if a change in the Bionic stemcell to mounts is causing the agent to report erroneous data.  A `df - h` does show the root partition of the stemcell being ~75% in use.  We have not moved forward to the latest Bionic stemcell `1.25` yet as we pinned `1.22` for now until we migrate our whole environment over to Bionic.  I do see from the Bionic release notes that the `bosh-agent` between `1.22` and `1.25` has not changed and is `2.346.0` for both stemcells: 

https://github.com/cloudfoundry/bosh-linux-stemcell-builder/releases/tag/ubuntu-bionic%2Fv1.25

https://github.com/cloudfoundry/bosh-linux-stemcell-builder/releases/tag/ubuntu-bionic%2Fv1.22

Thanks

bgandon:
This PR fixes #245, introducing a call to `IsMountPoint()` in order to exclude from vitals the existing directories where no disk is actually mounted.
The problem is a bit tricky because Windows and dummy platforms don't have any disk mounter. The proposed solution here, is based on an optional `diskMounter` dependency in the vitals service, that can be `nil`.
Unit tests for the vitals service have also been refactored a bit, in order to be clearer and adopt standard easy-to-read structure.
Cheers

### DIFF
--- a/agent/bootstrap_test.go
+++ b/agent/bootstrap_test.go
@@ -1116,7 +1116,7 @@ var _ = Describe("bootstrap", func() {
 
 				sigarCollector := boshsigar.NewSigarStatsCollector(&sigar.ConcreteSigar{})
 
-				vitalsService := boshvitals.NewService(sigarCollector, dirProvider)
+				vitalsService := boshvitals.NewService(sigarCollector, dirProvider, mounter)
 
 				ipResolver := boship.NewResolver(boship.NetworkInterfaceToAddrsFunc)
 
--- a/platform/dummy_platform.go
+++ b/platform/dummy_platform.go
@@ -69,7 +69,7 @@ func NewDummyPlatform(
 		copier:             boshcmd.NewGenericCpCopier(fs, logger),
 		dirProvider:        dirProvider,
 		devicePathResolver: devicePathResolver,
-		vitalsService:      boshvitals.NewService(collector, dirProvider),
+		vitalsService:      boshvitals.NewService(collector, dirProvider, nil),
 		certManager:        boshcert.NewDummyCertManager(fs, cmdRunner, 0, logger),
 		logger:             logger,
 		auditLogger:        auditLogger,
--- a/platform/linux_platform_test.go
+++ b/platform/linux_platform_test.go
@@ -80,7 +80,6 @@ var _ = Describe("LinuxPlatform", func() {
 		cdutil = fakecdrom.NewFakeCDUtil()
 		compressor = boshcmd.NewTarballCompressor(cmdRunner, fs)
 		copier = boshcmd.NewGenericCpCopier(fs, logger)
-		vitalsService = boshvitals.NewService(collector, dirProvider)
 		netManager = &fakenet.FakeManager{}
 		certManager = new(certfakes.FakeManager)
 		monitRetryStrategy = fakeretry.NewFakeRetryStrategy()
@@ -122,6 +121,8 @@ var _ = Describe("LinuxPlatform", func() {
 
 		diskUtil = fakedisk.NewFakeDiskUtil()
 		diskManager.GetUtilReturns(diskUtil)
+
+		vitalsService = boshvitals.NewService(collector, dirProvider, mounter)
 	})
 
 	JustBeforeEach(func() {
--- a/platform/provider.go
+++ b/platform/provider.go
@@ -72,7 +72,7 @@ func NewProvider(logger boshlog.Logger, dirProvider boshdirs.Provider, statsColl
 	// Kick of stats collection as soon as possible
 	statsCollector.StartCollecting(SigarStatsCollectionInterval, nil)
 
-	vitalsService := boshvitals.NewService(statsCollector, dirProvider)
+	vitalsService := boshvitals.NewService(statsCollector, dirProvider, linuxDiskManager.GetMounter())
 
 	ipResolver := boship.NewResolver(boship.NetworkInterfaceToAddrsFunc)
 
--- a/platform/vitals/service.go
+++ b/platform/vitals/service.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/cloudfoundry/gosigar"
 
+	boshdisk "github.com/cloudfoundry/bosh-agent/platform/disk"
 	boshstats "github.com/cloudfoundry/bosh-agent/platform/stats"
 	boshdirs "github.com/cloudfoundry/bosh-agent/settings/directories"
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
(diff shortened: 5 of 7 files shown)


jpalermo:
Hey @bgandon,

We did some investigation into the issue too. We are not seeing the problem with either a freshly deployed bionic or xenial stemcell. But we do see the problem if we attach a persistent disk, and then remove it, or if we simply create `/var/vcap/store` as a directory on the ephemeral disk.

The proposed change, does resolve the problem, although it also introduces a lot of additional complexity.

What are your thoughts on a smaller change to [UnmountPersistentDisk](https://github.com/cloudfoundry/bosh-agent/blob/d80ddb1f2e226b129b028e2156f99e31f4a9ee38/platform/linux_platform.go#L1237) such that when it unmounts the disk from `/var/vcap/store`, it also removes `/var/vcap/store` from the disk.

This should resolve the issue we're seeing in our testing, but your proposed solution solves an additional use case, which is a bosh release manually creating `/var/vcap/store` and putting files in there.

We're not totally sure if that is a supported use case or not. It does sound like maybe what is happening in the original issue #245 since they said they did a `recreate`, which if you don't have a persistent disk, should leave you without a `/var/vcap/store` since that should only be created by the agent when mounting a disk.

Thoughts?

jpalermo:
CC: @ChrisMcGowan if you have thoughts on the above investigation too

bgandon:
> _We are not seeing the problem with either a freshly deployed bionic or xenial stemcell._

Yes indeed, because the `/var/vcap/store` directory is not present of the base VM image from the Stemcell.

> _your proposed solution solves an additional use case, which is a bosh release manually creating /var/vcap/store and putting files in there._

Yes indeed, and this is common place in the vast majority of Bosh jobs. Sometimes even if they don't need persistent storage, because the Release Authors have just used standard one-fit-all boilerplate Bash code.
See for example this template from S&W `bosh-gen`, which has been very popular:
https://github.com/orange-cloudfoundry/cassandra-boshrelease/blob/master/jobs/cassandra/templates/helpers/ctl_setup.sh#L37-L44

So, we need to consider here that there's nothing that prevents any Bosh jobs from creating their own data into `/var/vcap/store` at any time.

Would we only remove that directory when unmounting the disk, then we wouldn't address the actual issue. And worse than that, the problem would arise in even more surprising ways for our users. And surprise is exactly what we try to avoid. Plus, we need not adding more fragile logic in the Bosh Agent code base, and instead we need to address precisely the root causes for the problems we try to solve.

Obviously, the solution I'm contributing here is not as simple as one could dream of. I realized that quite soon, while scratching my head (quite hard) to find the most reasonable changes in terms of correctness and best fit into the object design of the code base. But this also explains why the issue has kept unaddressed for so many years.

The basic problem we're not supposed to solve here, is that the Windows object hierarchy is different from the one for Linux in many ways. Thus designing a solution cannot be simple. But here, we don't fall into the trap of fixing this design inconsistency.

rkoster:
Thanks @bgandon! LGTM

bgandon:
Great, thank you Ruben! 👍

rkoster:
@bgandon this change caused the following CI failure in our windows unit tests:
```
------------------------------
+ Failure [0.006 seconds]
Vitals service when no persistent disk is mounted [It] does not return vitals for persistent disk 
C:/var/vcap/data/worker/work/containers/00000hc3qjn/tmp/build/1944c6a2/bosh-agent/platform/vitals/service_test.go:163

  
  	Error Trace:	service_test.go:167
  	            				runner.go:113
  	            				runner.go:64
  	            				it_node.go:26
  	            				spec.go:215
  	            				spec.go:138
  	            				spec_runner.go:200
  	            				spec_runner.go:170
  	            				spec_runner.go:66
  	            				suite.go:79
  	            				ginkgo_dsl.go:238
  	            				ginkgo_dsl.go:213
  	            				vitals_suite_test.go:12
  	Error:      	Should be false
  	Test:       	Vitals service when no persistent disk is mounted does not return vitals for persistent disk
  	Messages:   	Expected object with keys "ephemeral, persistent, system" to not have key "persistent"
  

  C:/var/vcap/data/worker/work/containers/00000hc3qjn/tmp/build/1944c6a2/bosh-agent/vendor/github.com/cloudfoundry/bosh-utils/assert/json_matchers.go:46
------------------------------


Summarizing 1 Failure:

[Fail] Vitals service when no persistent disk is mounted [It] does not return vitals for persistent disk 
C:/var/vcap/data/worker/work/containers/00000hc3qjn/tmp/build/1944c6a2/bosh-agent/vendor/github.com/cloudfoundry/bosh-utils/assert/json_matchers.go:46
```

Would you be so kind and fix this test?

jpalermo:
We managed to fix the failing test with these 2 commits:
https://github.com/cloudfoundry/bosh-agent/commit/2813e56f162a819241b7a284c1b7ed611645740d
https://github.com/cloudfoundry/bosh-agent/commit/7f7381d0189f8760a39fe01b7ddde96ee62a0b6a

Just a windows path issue

PR #247, opened 2021-09-28
Enable UseMTU in systemd networking configuration

jossy:
Without UseMTU, systemd will not properly use the MTU
value that is received over DHCP on Ubuntu Bionic.
The change should be compatible with previous stemcells,
since apparently the default behavior of systemd-networkd
was changed between Xenial and Bionic.

Diff of the relevant man page section (Xenial and Bionic stemcells):

> UseMTU=
>     When true, the interface maximum transmission unit from the DHCP server will be used on the current link. {+If MTUBytes= is set, then this setting is ignored.+} Defaults to [-true.-]{+false.+}

### DIFF
--- a/platform/net/ubuntu_net_manager.go
+++ b/platform/net/ubuntu_net_manager.go
@@ -604,6 +604,7 @@ func (net UbuntuNetManager) writeDynamicInterfaceConfiguration(config DHCPInterf
 	// DHCP Section
 	dhcpSection := &ini.Section{Name: "DHCP"}
 	dhcpSection.AddKey("UseDomains", "yes")
+	dhcpSection.AddKey("UseMTU", "yes")
 	file.AppendSection(dhcpSection)
 
 	// Route Sections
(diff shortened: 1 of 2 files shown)


jossy:
:point_up: We stumbled upon this problem in our (production) Concourse deployment in GCP btw, where the default MTU size in VPC networks is 1460 in contrast to the standard 1500 on Linux machines:
https://cloud.google.com/vpc/docs/vpc#mtu

jossy:
To give a bit more context: 

At least on Google Cloud the wrong MTU wouldn't normally cause problems, because there are mechanisms in place to tolerate MTU mismatches. I assume that's why this went undetected for quite some time.

The problem only became apparent when Google Cloud had an incident which broke that mechanisms for a few days:
https://status.cloud.google.com/incidents/QSirAFiyN5yMeeE6GNxq

We run both Xenial and Bionic stemcells in production and we noticed that only deployments using Bionic were affected by the Google incident. It took us a while to figure out what caused the networking issues we saw until we were finally able to track down the incorrect MTU setting.

ramonskie:
Seems as a good idea to enable this. Thank for the pr @jossy

PR #248, opened 2021-09-28
Feature: Adding support for NATS/Blobstore cert rotation without recreating vms

lnguyen:
Currently rotating nats or blobstore certs, vms need to be recreated so BOSH agent can get the new certs. 

This was proposed https://github.com/cloudfoundry/bosh/discussions/2309 and is the PR for the changes on the agent side


agent/action/update_settings_test.go
```diff
@@ -39,35 +40,33 @@ var _ = Describe("UpdateSettings", func() {
...
 			action.Run(newUpdateSettings)
-			expectedPath := filepath.Join(platform.GetDirProvider().BoshDir(), "update_settings.json")
-			exists := platform.GetFs().FileExists(expectedPath)
-			Expect(exists).To(Equal(true))
+			Expect(settingsService.SaveUpdateSettingsCallCount).To(Equal(1))
 		})
 	})
 
-	Context("when it cannot write the update settings file", func() {
+	Context("when it save update settings", func() {
```
bgandon, line 67: The test setup is basically the same: a situation where writing the settings fails.
Why should we change the test context description here?

bgandon (commented):
Nice work.

Just a small remark, it would be nice to have separate commits for new/updated dependencies vs. new code related to the new feature. Untangling those with `git rebase -i` for example would improve the diff clarity.

agent/action/update_settings_test.go
jpalermo: Looks like a bit of a typo, probably should be something like "when it cannot save update settings".

I'm guessing the change was because the UpdateSettings Action used to be responsible for writing the file to disk itself, but now that responsibility is handled by the settingsService. So these tests are now concerned with the settingsService returning an error (any error), and less concerned with the specifics of a particular error when writing a file.

settings/update_settings.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (updateSettings *UpdateSettings) MergeSettings(newSettings UpdateSettings) bool {
+	var restartNeeded bool
+
+	updateSettings.TrustedCerts = newSettings.TrustedCerts
+	updateSettings.DiskAssociations = newSettings.DiskAssociations
+
+	if !reflect.DeepEqual(newSettings.Mbus, updateSettings.Mbus) && !reflect.DeepEqual(newSettings.Mbus, MBus{}) {
+		updateSettings.Mbus = newSettings.Mbus
+		restartNeeded = true
```
beyhan, line 20: We could rename `restartNeeded` to something like `changed` because `UpdateSettings` shouldn't know anything about restart. A really minor thing.

jpalermo: Good call, we made it a big more verbose since `changed` could have been misleading since the trusted certs or disk associations could have changed and this value would have been `false` still.

PR #249, opened 2021-11-09
Add AWS IMDSv2 support

jpalermo:
Adds support for a new tokenPath field in agent.json

When this field is present, the agent will make a PUT request to that path to fetch a token before each AWS Instance Metadata Service call, and will include that token in a header when calling IMDS.

When the tokenPath is not specified, the agent behaves as before and uses IMDSv1.

It should also fall back to IMDSv1 if IMDSv2 is not available in the region. We're not sure if this exists, but we ran into similar problems recently with availability of gp3 disks in some regions.

This functionality being used is dependent on an update to the `agent.json` in the bosh linux stemcell builder which will be coming soon.

### DIFF
--- a/infrastructure/http_metadata_service.go
+++ b/infrastructure/http_metadata_service.go
@@ -23,6 +23,7 @@ type HTTPMetadataService struct {
 	userdataPath    string
 	instanceIDPath  string
 	sshKeysPath     string
+	tokenPath       string
 	resolver        DNSResolver
 	platform        boshplat.Platform
 	logTag          string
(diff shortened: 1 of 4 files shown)


ragaskar:
Is it possible to get some review eyes on this? We're blocking some other work on this merge occurring. Thanks! :)

beyhan (approved):

rkoster:
Thanks @jpalermo

ragaskar:
thank you!

PR #254, opened 2021-12-08
This moves /var/opt to /var/vcap/data/root_opt

jpalermo:
Similar to what we do with /var/log, we've seen issues where
people are installing 3rd party software on stemcells using automation.

For example, the Microsoft OMI Agent.

This software gets installed into /var/opt and logs into /var/opt/omi/log.
This ends up filling up the root disk eventually and VMs need
to be recreated. This software has no configuration options to log
into /var/log (and it seems that logging into /var/opt/THING/log
may be the correct linux spot to log to)

While this is not the suggested way to install software
onto a stemcell, this type of software is often put there
by people not involved with Bosh lifecycles.

Signed-off-by: Joseph Palermo <jpalermo@pivotal.io>

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -1068,6 +1068,9 @@ func (p linux) SetupOptDir() error {
 
 	boshRootOptDirPath := path.Join(p.dirProvider.DataDir(), "root_opt")
 	err := p.fs.MkdirAll(boshRootOptDirPath, userRootOptDirPermissions)
+	if err != nil {
+		return bosherr.WrapError(err, "Creating root opt dir")
+	}
 
 	_, _, _, err = p.cmdRunner.RunCommand("chown", "root:root", boshRootOptDirPath)
 	if err != nil {
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -40,6 +40,7 @@ const (
 	userBaseDirPermissions    = os.FileMode(0755)
 	disksDirPermissions       = os.FileMode(0755)
 	userRootLogDirPermissions = os.FileMode(0775)
+	userRootOptDirPermissions = os.FileMode(0755)
 	tmpDirPermissions         = os.FileMode(0755) // 0755 to make sure that vcap user can use new temp dir
 	blobsDirPermissions       = os.FileMode(0700)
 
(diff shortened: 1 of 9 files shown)


jpalermo:
We weren't thrilled about adding a new `SetupOptDir` method to the Platform Interface as this method doesn't really apply to Windows platforms, but it seemed better than the other options for testability.

From a design perspective, the `SetupLogDir` and `SetupOptDir` functionality seems like it should actually be part of `SetupRootDisk`, but there are actually a lot of other things that also could be bucketed into that same category, and since `SetupRootDisk` is doing a bunch of partitioning and resizing we didn't want to complicate the tests for that with additional factors.

ramonskie (approved):
lgtm

jpalermo:
Force pushed an update. There were some changes needed for the integration test helpers to clean up the mount points between tests.

platform/linux_platform.go
```diff
@@ -1062,6 +1063,25 @@ func (p linux) SetupLogDir() error {
 	return nil
 }
 
+func (p linux) SetupOptDir() error {
+	optDir := "/var/opt"
+
+	boshRootOptDirPath := path.Join(p.dirProvider.DataDir(), "root_opt")
+	err := p.fs.MkdirAll(boshRootOptDirPath, userRootOptDirPermissions)
```
beyhan, line 1070: I have two questions here:

1. is it by intention to ignore the error? Other places react on this e.g. https://github.com/cloudfoundry/bosh-agent/blob/8d18e4294e235edeb7f7fefc2f081ce006f57a71/platform/linux_platform.go#L986-L988
2. if the `boshRootOptDirPath` already exists `MkdirAll` won't do anything (as described in https://github.com/cloudfoundry/bosh-agent/blob/8d18e4294e235edeb7f7fefc2f081ce006f57a71/vendor/github.com/charlievieth/fs/fs.go#L90-L97). That means the `userRootOptDirPermissions` won't be applied. Should we do an explicit `chmod`?

jpalermo: Good catch, we apparently totally failed in our copy/paste from the /var/log code and didn't get the error handling. We just added that in a new commit.

We did look at the MkdirAll originally and opted not to do a `chmod` because we should be the only people making that path, so if it was made previously, it should have been made with the correct permissions.

We went back and took another look at it because we thought it might be possible for the `MkdirAll` to fail in the middle; make the director, but then the agent crashes before it can `chmod` it, and now the VM is in a bad state. But it seems like that's not possible, the mkdir syscall (way down in the call chain) takes both the directory name to create, and the permissions to give that directory, so I'm guessing it's an atomic operation within Linux.

beyhan: It sounds reasonable. @jpalermo  Thank you for the detailed consideration.

beyhan (approved):

PR #256, opened 2022-01-19
When mounting /var/opt to the ephemeral disk, also mount /opt

jpalermo:
We've heard reports of additional agents being installed into `/opt` and filling up the root disk. `/opt` is empty by default so this should be a safe operation.

We changed the mount point for `/var/opt` from `/var/vcap/data/root_opt` to `/var/vcap/data/root_var_opt` to avoid name confusion. Since this is all on the ephemeral disk it shouldn't cause breakage when those paths change.

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -1082,6 +1082,8 @@ func (p linux) SetupOptDir() error {
 		return bosherr.WrapError(err, "Chowning root var opt dir")
 	}
 
+	//Mount our /var/opt bind mount without the 'noexec' option. Binaries are
+	//often in subdirectories of /var/opt, and folks expect to be able to execute them.
 	err = p.bindMountDir(boshRootVarOptDirPath, varOptDir, true)
 	if err != nil {
 		return err
@@ -1100,6 +1102,8 @@ func (p linux) SetupOptDir() error {
 		return bosherr.WrapError(err, "Chowning root opt dir")
 	}
 
+	//Mount our /opt bind mount without the 'noexec' option. Binaries are
+	//often in subdirectories of /opt, and folks expect to be able to execute them.
 	err = p.bindMountDir(boshRootOptDirPath, optDir, true)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


lnguyen (approved):
LGTM

jpalermo:
We got a report in CF Slack that the new bind mounting of `/var/opt` causes problems because we mount it with `noexec`. Just added another commit to this PR that fixes this change in behavior for both `/var/opt` and `/opt`

Context: https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1642679288035300

platform/linux_platform.go
```diff
@@ -1069,10 +1069,28 @@ func (p linux) SetupLogDir() error {
...
+	if err != nil {
+		return bosherr.WrapError(err, "Creating root var opt dir")
+	}
+
+	_, _, _, err = p.cmdRunner.RunCommand("chown", "root:root", boshRootVarOptDirPath)
+	if err != nil {
+		return bosherr.WrapError(err, "Chowning root var opt dir")
+	}
+
+	err = p.bindMountDir(boshRootVarOptDirPath, varOptDir, true)
```
bgandon, line 1085: All right for the issue with `noexec` that has been reported in Slack. But we should document here _why_ we do this, so that future maintainers after us will know _how_ they can (ideally) re-test for the workaround still being necessary, and know _when_ they can remove the workaround code.

bgandon (changes requested):
Globally looks good to me. A bit of commenting non-obvious workarounds would be nice 👍

platform/linux_platform.go
klakin-pivotal: @bgandon What do you think of this comment and the one further down the file?

klakin-pivotal: (If you like them, I can squash this into the previous commit and force push.)

bgandon (approved):

platform/linux_platform.go
bgandon: It's all right yes 👍

rkoster:
Thanks! @jpalermo

PR #258, opened 2022-02-08
Retry mechanism for http_blob_provider

OliverMautschke:
We experience intermittently failing Get request to fetch the blobs from an external blobstore due to network issues. To make the Get mechanism more resilient we introduce a retry strategy.

### DIFF
--- a/agent/httpblobprovider/blobstore_delegator/blobstore_delegator.go
+++ b/agent/httpblobprovider/blobstore_delegator/blobstore_delegator.go
@@ -2,6 +2,10 @@ package blobstore_delegator
 
 import (
 	"fmt"
+	bosherr "github.com/cloudfoundry/bosh-utils/errors"
+	boshlog "github.com/cloudfoundry/bosh-utils/logger"
+	boshretry "github.com/cloudfoundry/bosh-utils/retrystrategy"
+	"time"
 
 	httpblobprovider "github.com/cloudfoundry/bosh-agent/agent/httpblobprovider"
 	"github.com/cloudfoundry/bosh-utils/blobstore"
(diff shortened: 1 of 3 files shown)


rkoster:
@OliverMautschke Thanks for creating this PR. Could you give a bit more context on when this happens, and what the results are? I would assume this could result in a failed deployment, bonus points for bosh task log so people who encounter the same issue can find this PR.

In general when retrying certain operations, other places in the codebase use the retrystragegy package from the bosh-utils repo. In particular I think the following method would be useful here: [NewAttemptRetryStrategy](https://pkg.go.dev/github.com/cloudfoundry/bosh-utils@v0.0.300/retrystrategy#NewAttemptRetryStrategy).

ramonskie (approved):
LGTM

rkoster:
Thanks! @OliverMautschke

friegger:
To help others identify the problem, the error would look like the following:
```
Task 3032457 | 17:17:41 | Updating instance <instance_name>: <instance>/4e634857-1f78-4c3e-9a50-352bae34d39a (0) (canary) (00:00:38)
                        L Error: Action Failed get_task: Task 31ec5186-d5de-45de-51aa-4fb7cc13cb55 result: Preparing apply spec: Preparing package <package_name>: Fetching package blob: Excuting GET request: Get "https://<bucket_name>.s3.amazonaws.com/9f7aa6ba-2036-4d3c-a920-bcc05c2c1ee7?<signed_url_params>: dial tcp <ip>:443: i/o timeout
```
Sometimes it is also `EOF` or `net/http: TLS handshake timeout`. The Director uses signed blobstore urls.

PR #259, opened 2022-02-15
Add interfaces and implementations for Nats FW Win

nouseforaname:
Add code to the agent to secure the remote mbus that
bosh provides. Provide a mechanism to restrict workloads
on the VM other than the agent itself to connect to BOSHs
mbus.

On VM creation bosh provides the Agent on the VM with creds
to connect to Nats. Workloads that live on the vm could
extract these credentials and use them to connect to nats and
extract further credentials that are being sent over nats once
the rendering of jobs takes place.

This creates a FW rule within the agent bootstrap to limit which
binaries can create outgoing requests to the nats endpoint.

The created FW rules are ephemeral and exist as long as the agent
process runs. This avoids the necessity for cleaning up FW rules if
the remote address of the nats should change. If a process starts
before the rule exists, the rule will still apply once it exists.

Updated tests relying on SetupNetwork calls with new method signature

https://www.pivotaltracker.com/story/show/181114075


beyhan:
Have you consider to make this optional? I didn't look into the implementation but if the probability to have issues is high I would say that it make sense to make it optional.

beyhan:
After discussions about this with @nouseforaname we decided to drop the option to make this optional. This change will be introduced with a new stemcell, which can be easily reverted in case of unexpected issue which we don't expect.

platform/windows_platform.go
```diff
@@ -28,7 +28,7 @@ import (
 	boshuuid "github.com/cloudfoundry/bosh-utils/uuid"
 )
 
-//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o fakes/fake_windows_disk_manager.go . WindowsDiskManager
+//go:generate go run github.com/maxbrunsfeld//v6 -o fakes/fake_windows_disk_manager.go . WindowsDiskManager
```
jpalermo, line 31: Not sure what happened here, but this change doesn't seem correct. Not sure if it was a typo, or if something weird is going on with `go generate`

jpalermo (commented):
Looks good otherwise. Thanks!

platform/windows_platform.go
nouseforaname: I guess that was my VSCode. Regenerating everything worked afterwards. Not that familiar with counterfeiter and will believe you and revert this line if you think that is the way to go.

ramonskie (approved):
LGTM.

as any errors would definitely be picked up in bats testing when creating a new stemcell.
i would inform both linux and windows stemcell teams that this change is coming up.

platform/windows_platform.go
jpalermo: Yeah, running `go generate` on my machine without the `counterfeiter` segment results in:
```
cannot find package "." in:
	/Users/jpalermo/workspace/bosh-agent/vendor/github.com/maxbrunsfeld/v6
windows_platform.go:31: running "go": exit status 1

Process finished with exit code 1
```
I also had an `inconsistent vendoring` error when first trying to run `go generate`, you might need to `go mod tidy` & `go mod vendor`.

nouseforaname: I reverted  the line and did a go mod tidy && vendor. Shoul'd be fixed now. Can you test on your end if it works now?

jpalermo:
Looks good!

PR #264, opened 2022-02-24
Implement nats firewall for linux. add integration test (do not merge before PR-259)

nouseforaname:
This commit adds the SetupFirewall implemention for Linux agents.
DO not merge before the windows counterpart is merged since that [PR ](https://github.com/cloudfoundry/bosh-agent/pull/259) introduces the interfaces. Once the other PR is approved and merged I will rebase this branch onto main.

unit tests are passing, integration tests are passing (ran them multiple times in a row to rule out flakyness because of the
complicated setup)

`functionality`
why return early if mbusUrl is ""? 
Because on some IaaSs we require calling SetupNetwork multiple times.
Specifically [here](https://github.com/cloudfoundry/bosh-agent/blob/9339c991d3abadfcc9e2e59fe4646f1f3440419a/infrastructure/http_metadata_service.go#L308)

`test changes`
I had to do quite a bit of updating for the setup integration tests.
The background is that once the registry related tests ran, they rewrote
the original json files in /var/vcap/bosh and starting an agent without
the registry failed. Instead of making it work with the registry I opted 
to take a backup of the original settings instead. The reasoning is that 
we will soon remove all the registry code anyway and will have to rewrite
the integration tests without requiring the registry for setups. 

The iptables rules that are added within the suite setup need to be there
to support the agent tunneling. The ssh tunnel uses linux ephemeral ports
to forward the requests on the target host. Since each linux system can
have different settings, I used to output of `cat /proc/sys/net/ipv4/ip_local_port_range` 
on a bionic 1.61 VM. 
Since the added integration test does not rely on tunneling requests to the agent, that
rule is dropped in this tests before each and readded once it's done. 

Additionally this PR updates the RunCommand method to actually
return stderr if a command fails so we can check for expected linux
error output.

Tracker Story: #180698114


lnguyen (commented):
Overall looks good to me, do we forsee any problems with differences between xenial, bionic, and jammy?

platform/net/firewall_provider_linux.go
```diff
@@ -3,7 +3,111 @@
...
+		Network: &specs.LinuxNetwork{
+			ClassID: &isolationClassID,
+		},
+	})
+	if err != nil {
+		return bosherr.WrapError(err, "Error setting up cgroups for nats api access")
+	}
 
-	return nil
+	err = natsAPICgroup.AddProc(uint64(os.Getpid()), cgroups.NetCLS)
```
beyhan, line 111: The `classid` definition was done before the `iptable` rules are set in the script based version. See https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/219/files#diff-9d395104639f75c39a4b71d21f0da02530a19ad79d52dd26e824c43c35e9809cR9. Is it important whether the `classid` definition is configured before or after the `iptable` rules?

platform/net/ubuntu_net_manager.go
```diff
@@ -89,20 +89,24 @@ func (net UbuntuNetManager) SetupIPv6(config boshsettings.IPv6, stopCh <-chan st
...
 func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, mbus string, errCh chan error) error {
 	if networks.IsPreconfigured() {
-		err := net.setupFirewall(mbus)
+
+		// Note in this case IPs are not broadcast
+		dnsNetwork, _ := networks.DefaultNetworkFor("dns")
+		net.logger.Info(UbuntuNetManagerLogTag, "Successfully set up outgoing nats api firewall")
+		err := net.writeResolvConf(dnsNetwork.DNS)
+		if err != nil {
+			return err
```
beyhan, line 98: Is the DNS configuration order changed because of the `hostname` use case for NATs?

platform/net/ubuntu_net_manager.go
```diff
@@ -89,20 +89,24 @@ func (net UbuntuNetManager) SetupIPv6(config boshsettings.IPv6, stopCh <-chan st
 
 func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, mbus string, errCh chan error) error {
 	if networks.IsPreconfigured() {
-		err := net.setupFirewall(mbus)
+
+		// Note in this case IPs are not broadcast
+		dnsNetwork, _ := networks.DefaultNetworkFor("dns")
+		net.logger.Info(UbuntuNetManagerLogTag, "Successfully set up outgoing nats api firewall")
```
beyhan, line 95: Is this log message really correct? The firewall rules are not set up at this point.

platform/net/ubuntu_net_manager.go
```diff
@@ -164,20 +166,12 @@ func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, mbus
...
-
-func (net UbuntuNetManager) setupFirewall(mbus string) error {
-	if mbus == "" {
-		net.logger.Info("NetworkSetup", "Skipping adding Firewall for outgoing nats. Mbus url is empty")
-		return nil
+		return bosherr.WrapError(err, "Setting up nats firewall")
 	}
-	net.logger.Info("NetworkSetup", "Adding Firewall not implemented on")
+	net.logger.Info(UbuntuNetManagerLogTag, "Successfully set up outgoing nats api firewall")
+	go net.addressBroadcaster.BroadcastMACAddresses(append(staticAddressesWithoutVirtual, dynamicAddresses...))
```
beyhan, line 174: Why we moved the IPs broadcast to the end? It was before the `setupFirewall` call https://github.com/cloudfoundry/bosh-agent/pull/264/commits/7d56f583f8806b1b45d179c834e03e9deba94b0c#diff-010e8a578b784cca80730f6e56600727eede1f1fafdd186f718872bff04841e6L167

integration/integration_suite_test.go
```diff
@@ -28,15 +29,21 @@ func TestIntegration(t *testing.T) {
 
 		err = testEnvironment.StopAgent()
 		Expect(err).ToNot(HaveOccurred())
+		//create a backup of original settings for nats FW tests
+		_, err = testEnvironment.RunCommand("sudo sh -c \"mkdir -p /settings-backup && cp /var/vcap/bosh/*.json /settings-backup/ \" ")
+		Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
nouseforaname: without them being copied back into the test setup, those files wouldn't interfere with future tests.
I can extract it to another commit if required but I don't see much benefit as we have to rewrite all tests anyway. I;ll add some info about these lines into the tracker story.

platform/net/firewall_provider_linux.go
```diff
@@ -3,7 +3,111 @@
...
+		"--dport", "2822",
+		"-m", "cgroup",
+		"--cgroup", fmt.Sprintf("%v", natsIsolationClassID),
+		"-j", "ACCEPT",
+	)
+	if err != nil {
+		return bosherr.WrapError(err, "Iptables Error checking for monit rule")
+	}
+	if !exists {
+		err = ipt.Insert("mangle", "POSTROUTING", 1,
```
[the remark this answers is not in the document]
nouseforaname: we wan't to make sure that it will be before the already existing monit drop rule. that is still added via the pre start script in the stemcell. iptables won't evaluate after the first match, so our allow for the bosh agent needs to be inserted.

platform/net/firewall_provider_linux.go
nouseforaname: not to my understanding. basically my understanding of the way this works isL
net_cls tags network packets with the classid for processes in a cgroup. the iptables extension just looks at that tag. so if the cgroup and the associated class id don't exist yet, nothing can have that tag. I don't think that either needs to come first. it's just about when traffic filtering starts.

platform/net/ubuntu_net_manager.go
nouseforaname: I have just put skeleton code where I assumed I'd have to call the iptables logic when I created the windows pr. I noticed that dns probably should come first and moved it when I actually implemented it.

platform/net/ubuntu_net_manager.go
nouseforaname: same as above. I just put in skelethon code when the windows counterpart was written.

nouseforaname:
> Overall looks good to me, do we forsee any problems with differences between xenial, bionic, and jammy?

I dont think so. we're relying on cgroups v1 and iptables. iptables interface doesn't change much and cgroupsv1 are not dropped with the release of v2 they can coexist.

platform/net/firewall_provider_linux.go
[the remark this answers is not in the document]
beyhan: Ok, I see. Thanks!

platform/net/firewall_provider_linux.go
beyhan: I'm also not so deep into this  that is why I wanted to have a discussion on this.

rkoster:
Thanks! @nouseforaname

PR #261, opened 2022-02-23
Fix panic in sfdiskPartitioner.GetPartitions from extra whitespace in…

Justin-W:
… sfdisk output. (Fixes #260.)

NOTE: Prior to this change, the extractPartitionPathAndType func would panic with "panic: runtime error: index out of range [-1]" if there were any whitespace-only or blank lines in the STDOUT output from the `sfdisk -d` command. For example, a trailing blank line.

PR #266, opened 2022-02-25
Replace the `sfdisk -uM` command with a cross-platform compatible...

### LINKED ISSUE #265: sfdisk -uM causes partitioning to fail on some linux platforms
The agent currently performs disk partitioning using the `sfdisk -uM` command. The `-uM` option is unsupported on many linux platforms, and causes partitioning to fail on stemcells for such platforms.

Here is an example of such a failure.

```
2022-02-24_01:46:47.33033 [Cmd Runner] 2022/02/24 01:46:47 DEBUG - Running command 'sfdisk -uM /dev/sdc'
2022-02-24_01:46:47.33307 [Cmd Runner] 2022/02/24 01:46:47 DEBUG - Stdout: 
2022-02-24_01:46:47.33309 [Cmd Runner] 2022/02/24 01:46:47 DEBUG - Stderr: sfdisk: unsupported unit 'M'
2022-02-24_01:46:47.33310 [Cmd Runner] 2022/02/24 01:46:47 DEBUG - Successful: false (1)
2022-02-24_01:46:47.33310 [SfdiskPartitioner] 2022/02/24 01:46:47 ERROR - Failed with an error: Running command: 'sfdisk -uM /dev/sdc', stdout: '', stderr: 'sfdisk: unsupported unit 'M'
2022-02-24_01:46:47.33311 ': exit status 1
```

This [sfdisk man page](https://man7.org/linux/man-pages/man8/sfdisk.8.html#OPTIONS) doesn't list `-uM` as a supported unit, and says that the `-u` option is deprecated even for the single `-uS` unit that is documented.

The agent should be modified to replace the `sfdisk -uM` command with something that is more widely supported across linux platforms.

Justin-W:
... equivalent. (Fixes #265.)

This change adds explicit "multiplicative suffixes" in the sfdisk input to specify the units of the partition sizes (instead of the `-uM` command option, since `-uM` appears to be unsupported on many platforms and `-u` seems to be deprecated on many platforms). This change should produce identical partitioning behavior (vs the previous implementation), using an `sfdisk` input format which is more widely supported across linux platforms.

NOTE: Prior to this change, the `sfdisk -uM` command failed (on some stemcells), with agent log output like the following.

> ... ERROR - Failed with an error: Running command: 'sfdisk -uM /dev/sdc', stdout: '', stderr: 'sfdisk: unsupported unit 'M'

See:
- https://man7.org/linux/man-pages/man8/sfdisk.8.html#OPTIONS
- https://man7.org/linux/man-pages/man8/sfdisk.8.html#INPUT_FORMATS
- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/considerations_in_adopting_rhel_8/file-systems-and-storage_considerations-in-adopting-rhel-8#removal-of-cylinder-head-sector-addressing-from-sfdisk-and-cfdisk_file-systems-and-storage

### DIFF
--- a/platform/disk/sfdisk_partitioner.go
+++ b/platform/disk/sfdisk_partitioner.go
@@ -46,7 +46,7 @@ func (p sfdiskPartitioner) Partition(devicePath string, partitions []Partition) 
 	sfdiskInput := ""
 	for index, partition := range partitions {
 		sfdiskPartitionType := sfdiskPartitionTypes[partition.Type]
-		partitionSize := fmt.Sprintf("%d", ConvertFromBytesToMb(partition.SizeInBytes))
+		partitionSize := fmt.Sprintf("%dMiB", ConvertFromBytesToMb(partition.SizeInBytes))
 
 		if index == len(partitions)-1 {
 			partitionSize = ""
@@ -56,7 +56,7 @@ func (p sfdiskPartitioner) Partition(devicePath string, partitions []Partition) 
 	}
 
 	partitionRetryable := boshretry.NewRetryable(func() (bool, error) {
-		_, _, _, err := p.cmdRunner.RunCommandWithInput(sfdiskInput, "sfdisk", "-uM", devicePath)
+		_, _, _, err := p.cmdRunner.RunCommandWithInput(sfdiskInput, "sfdisk", devicePath)
 		if err != nil {
 			p.logger.Error(p.logTag, "Failed with an error: %s", err)
 			return true, bosherr.WrapError(err, "Shelling out to sfdisk")
--- a/platform/disk/sfdisk_partitioner_test.go
+++ b/platform/disk/sfdisk_partitioner_test.go
@@ -110,7 +110,7 @@ var _ = Describe("sfdiskPartitioner", func() {
 		partitioner.Partition("/dev/sda", partitions)
 
 		Expect(1).To(Equal(len(runner.RunCommandsWithInput)))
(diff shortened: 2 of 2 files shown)


Justin-W:
FYI: This PR currently branches off of a commit from open PR #261, because the new test which was added in #261 also needed to be updated as a result of the code fixes for #265 in this PR.

If for some reason #261 gets rejected, I can revise this PR to fix #265 independent of the changes in #261.

Justin-W:
> FYI: This PR currently branches off of a commit from open PR #261, because the new test which was added in #261 also needed to be updated as a result of the code fixes for #265 in this PR.

I rebased onto `main` now that #261 was merged, so this PR no longer includes unrelated commits.

ramonskie (commented):
looks good to me.
the only thing that worries me as that some time ppl use sizes like `20_500` etc
so not sure how that is handles

Justin-W:
> looks good to me. the only thing that worries me as that some time ppl use sizes like `20_500` etc so not sure how that is handles

Since [ConvertFromBytesToMb](https://github.com/cloudfoundry/bosh-agent/blob/4c27097cfd2a27a41ddc9deacbcd8b11ca07388f/platform/disk/convert.go#L3) returns an int, my change should work with all returnable values. Sizes like `20_500` must already be converted during config parsing.

jpalermo (approved):
Looks good. Thanks!

PR #268, opened 2022-03-16
Return early if v1 cgroups are not mounted

nouseforaname:
on warden the nats fw feature breaks the agent start with

`2022-03-15_22:58:56.43738 [main] 2022/03/15 22:58:56 ERROR - Agent exited with error: Running bootstrap: Setting up networking: Setting up Nats Firewall: Error setting up cgroups for nats api access: cgroups: cgroup mountpoint does not exist`

add check if cgroups are mounted to the if that handles early returns.

### DIFF
--- a/platform/net/firewall_provider_linux.go
+++ b/platform/net/firewall_provider_linux.go
@@ -4,6 +4,7 @@
 package net
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	gonetURL "net/url"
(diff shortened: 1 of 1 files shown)


ramonskie (approved):
LGTM

platform/net/firewall_provider_linux.go
```diff
@@ -37,8 +37,10 @@ const (
 func SetupNatsFirewall(mbus string) error {
 	// return early if
 	// we get a https url for mbus. case for create-env
-	// we get an empty string. case for http_metadata_service (responsible to extract the agent-settings.json from the metadata endpoint)s
-	if mbus == "" || strings.HasPrefix(mbus, "https://") {
+	// we get an empty string. case for http_metadata_service (responsible to extract the agent-settings.json from the metadata endpoint)
+	// we find that v1cgroups are not mounted (warden stemcells)
+	availableV1Groups, _ := cgroups.V1()
```
cunnie, line 42: I'm a little uncomfortable tossing away the error; I've been bitten by that too many times. Maybe add an error check here?

Also, is there any testing we can easily do? It doesn't seem like we're using interfaces, so the answer is probably no, but it would be nice to have them.

nouseforaname: Yeah that makes  sense. Thanks for spotting this.

cunnie (approved):
Looks good to me!

PR #271, opened 2022-03-23
Feature/nats fw 2.268

nouseforaname:
Backported commits in main branch for:
- add updated interface to centos, rhel etc. netmanagers
- recreate fakes
- go mod tidy
- go mod vendor

ran unit-tests

https://github.com/cloudfoundry/bosh-agent/pull/259
```
    Add interfaces and implementations for Nats FW Win

    Add code to the agent to secure the remote mbus that
    bosh provides. Provide a mechanism to restrict workloads
    on the VM other than the agent itself to connect to BOSHs
    mbus.

    On VM creation bosh provides the Agent on the VM with creds
    to connect to Nats. Worklaoads that live on the vm could
    extract these credentials and use them to connect to nats and
    extract further credentials that are being sent over nats once
    the rendering of jobs takes place.

    This creates a FW rule within the agent bootstrap to limit which
    binaries can create outgoing requests to the nats endpoint.

    The created FW rules are ephemeral and exist as long as the agent
    process runs. This avoids the necessity for cleaning up FW rules if
    the remote address of the nats should change. If a process starts
    before the rule exists, the rule will still apply once it exists.

    Updated tests relying on SetupNetwork calls with new method signature
```

https://github.com/cloudfoundry/bosh-agent/pull/264

```
This commit adds the SetupFirewall implemention for Linux agents.

unit tests are passing, integration tests are passing (ran them multiple
times in a row to rule out flakyness because of the complicated setup)

functionality
why return early if mbusUrl is ""?
Because on some IaaSs we require calling SetupNetwork multiple times.
The first time it is called we cannot possible have the mbus url known and
therefore needed a way to make this idempotent. Specifically [example here](https://github.com/cloudfoundry/bosh-agent/blob/9339c991d3abadfcc9e2e59fe4646f1f3440419a/infrastructure/http_metadata_service.go#L308)

test changes
I had to do quite a bit of updating for the setup integration tests.
The background is that once the registry related tests ran, they rewrote
the original json [file](https://github.com/cloudfoundry/bosh-agent/blob/f974e36207e2105082f1ecf935f64974583a3283/integration/assets/config-drive-agent.json#L12)s in /var/vcap/bosh and starting an agent without
the registry failed. Instead of making it work with the registry I opted
to take a backup of the original settings instead. The reasoning is that
we will soon remove all the registry code anyway and will have to rewrite
the integration tests without requiring the registry for setups.

To make sure that the backup of the original settings (from the time the vm
is deployed) is only run once, I switched the BeforeSuite to SynchronizedBeforeSuite.
When writing and testing the test I noticed that in some cases the agent.json
in /var/vcap/bosh was overwritten with the contents of this file

Additionally this PR updates the RunCommand method to actually
return stderr if a command fails so we can check for expected linux
error output.

Tracker Story: #180698114
```

https://github.com/cloudfoundry/bosh-agent/pull/268

```
on warden the nats fw feature breaks the agent start with

2022-03-15_22:58:56.43738 [main] 2022/03/15 22:58:56 ERROR - Agent exited with error: Running bootstrap: Setting up networking: Setting up Nats Firewall: Error setting up cgroups for nats api access: cgroups: cgroup mountpoint does not exist

add check if cgroups are mounted to the if that handles early returns
```


rkoster (approved):
LGTM

PR #273, opened 2022-03-30
Add V6 iptables support

nouseforaname:
currently a stemcell build with the agent will not be able to successfully
deploy since the iptables setup is trying to add v6 addresses via the v4 
api (using the binary iptables through the iptables lib) but in case of a
v6 target we need to use `ip6tables` .

This adds logic based on dns lookups to figure out if the target is using v4
or v6 and switched the creation of the `ipt` object to use the according ipVersion.

### DIFF
--- a/platform/net/firewall_provider_linux.go
+++ b/platform/net/firewall_provider_linux.go
@@ -59,12 +59,31 @@ func SetupNatsFirewall(mbus string) error {
 	if err != nil {
 		return bosherr.WrapError(err, "Error getting Port")
 	}
-	ipt, err := iptables.New()
+
+	// Run the lookup for Host as it could be potentially a Hostname | IPv4 | IPv6
+	// the return for LookupIP will be a list of IP Addr and in case of the Input being an IP Addr,
+	// it will only contain one element with the Input IP
+	addr_array, err := net.LookupIP(host)
 	if err != nil {
-		return bosherr.WrapError(err, "Iptables Error")
+		return bosherr.WrapError(err, fmt.Sprintf("Error resolving mbus host: %v", host))
 	}
+
+	// default to V4
+	localhostIP := "127.0.0.1"
+	ipVersion := iptables.ProtocolIPv4
+
+	// Check if we're dealing with a V4 Target
+	if addr_array[0].To4() == nil {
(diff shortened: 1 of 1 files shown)


platform/net/firewall_provider_linux.go
```diff
@@ -59,12 +59,32 @@ func SetupNatsFirewall(mbus string) error {
...
+
+	// Check if we're dealing with a V4 Target
+	v4_addr := addr_array[0].To4()
+	if v4_addr == nil {
+		localhostIP = "0:0:0:0:0:0:0:1"
+		ipVersion = iptables.ProtocolIPv6
 	}
+	ipt, err := iptables.NewWithProtocol(ipVersion)
+	if err != nil {
+		return bosherr.WrapError(err, "Creting Iptables Error")
```
rkoster, line 83: Typo `Creting`

platform/net/firewall_provider_linux.go
```diff
@@ -59,12 +59,32 @@ func SetupNatsFirewall(mbus string) error {
...
+	// it will only contain one element with the Input IP
+	addr_array, err := net.LookupIP(host)
 	if err != nil {
-		return bosherr.WrapError(err, "Iptables Error")
+		return bosherr.WrapError(err, fmt.Sprintf("Error resolving mbus host: %v", host))
+	}
+
+	// default to V4
+	localhostIP := "127.0.0.1"
+	ipVersion := iptables.ProtocolIPv4
```
rkoster, line 73: This variable does not seem to be used

platform/net/firewall_provider_linux.go
```diff
@@ -59,12 +59,32 @@ func SetupNatsFirewall(mbus string) error {
...
+
+	// Check if we're dealing with a V4 Target
+	v4_addr := addr_array[0].To4()
+	if v4_addr == nil {
+		localhostIP = "0:0:0:0:0:0:0:1"
+		ipVersion = iptables.ProtocolIPv6
```
rkoster, line 79: This could be inlined on line 81

rkoster (changes requested):
Looks good overall just a few small changes.

PR #272, opened 2022-03-28
Switch to golangci-lint meta-lint tool

aramprice:
Currently [bin/test-unit](https://github.com/cloudfoundry/bosh-agent/blob/main/bin/test-unit#L47-L50) runs  `go run golang.org/x/lint/golint` via [bin/golint](https://github.com/cloudfoundry/bosh-agent/blob/main/bin/golint#L9).

The `golang.org/x/lint/golint` command has been [deprecated and frozen](https://github.com/golang/go/issues/38968) since 2020.

This commit switches to the meta-lint tool https://golangci-lint.run/

Additional Changes
- Fixs https://github.com/koalaman/shellcheck/ issues with scripts in `bin/`
- Removes dead code discovered by `golangci-lint`
- Change formatter used by the project to `goimports` as recommended by `golangci-lint`
  - `goimports` adds `import` formatting and is otherwise the same as `go fmt`

PR #274, opened 2022-03-31
Fix the `bin/test-bosh-integration` script's output to match its behavior

Justin-W:
This fixes an inconsistency between output vs behavior that was introduced by edfed12f as part of #272.


rkoster:
Thanks! @Justin-W

rkoster (approved):

PR #275, opened 2022-04-04
Add V6 iptables support

nouseforaname:
Currently Stemcells build with the agent will not be able to successfully
deploy in a V6 setup. The reason/background is that the Iptables code is trying 
to add V6 addresses via the Iptables v4 API.  

Add logic to figure out if the actual IP behind the host in `mbus_string`
is an IPv4 vs IPv6 address and handle instantiation of Iptables lib accordingly

Monit runs on V4 locally, even on a V6 VM thus the created rule is always V4.

### DIFF
--- a/platform/net/firewall_provider_linux.go
+++ b/platform/net/firewall_provider_linux.go
@@ -61,10 +61,21 @@ func SetupNatsFirewall(mbus string) error {
 	if err != nil {
 		return bosherr.WrapError(err, "Error getting Port")
 	}
+
+	// Run the lookup for Host as it could be potentially a Hostname | IPv4 | IPv6
+	// the return for LookupIP will be a list of IP Addr and in case of the Input being an IP Addr,
+	// it will only contain one element with the Input IP
+	addr_array, err := net.LookupIP(host)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


rkoster (approved):
LGTM

lnguyen (commented):
LGTM

lnguyen (approved):

platform/net/firewall_provider_linux.go
```diff
@@ -89,6 +100,18 @@ func SetupNatsFirewall(mbus string) error {
 			return bosherr.WrapError(err, "Iptables Error inserting for monit rule")
 		}
 	}
+
+	// For nats iptables rules we default to V4 unless below dns resolution gives us a V6 target
+	ipVersion := iptables.ProtocolIPv4
+	// Check if we're dealing with a V4 Target
```
beyhan, line 106: This comment is wrong or? We are checking wether we deal with IP v6 here.

nouseforaname: yeah we're checking if it's a V6 Address by trying to create a V4. If that fails, we have a V6. Since the check is actually trying to run To4() I thought it's less confusing.

beyhan (approved):

PR #277, opened 2022-04-07
Sync DNS nameserver changes to 'resolv.conf.d/head'

### LINKED ISSUE #240: DNS entries from cloud-config are not getting updated correctly post bosh deployment
We are experiencing an issue with the DNS configuration in BOSH provisioned VMs. We are deploying to AWS and have 'DNS resolution' enabled by default at the VPC level. We have our own DNS provider running inside this VPC, so as part of our 'network' definition in the cloud-config, we are specifying the DNS entries for our provider. For context, the VPC CIDR is - 10.128.0.0/18, the 'default' VPC DNS nameserver is 10.128.0.2 and the addresses for the DNS provider is 10.128.1.11, 10.128.2.12, 10.129.0.13 - which is what is defined in the cloud config. The stemcell we are using is the most recent version of ubuntu-xenial - https://bosh-aws-light-stemcells.s3-accelerate.amazonaws.com/621.99/light-bosh-stemcell-621.99-aws-xen-hvm-ubuntu-xenial-go_agent.tgz.

The idea of defining the DNS entries in the cloud-config is for the agent to override the default DNS in the VM with the one provided by the cloud-config, but that is not working as expected. Post VM provisioning and update by BOSH, we expect the resolv.conf to have the following entries:

```
nameserver 10.128.1.11
nameserver 10.128.2.12
nameserver 10.129.0.13
search us-west-2.compute.internal
```
but, it has (notice how the AWS default DNS server is on top of the list):

```
nameserver 10.128.0.2
nameserver 10.128.1.11
nameserver 10.128.2.12
search us-west-2.compute.internal
```
This is effecting service discovery for our applications. Now looking through the code, and the agent logs, I can see that the agent calls the 'resolvconf -u' command directly to update the DNS entries, which basically is reading the 'base' file (which is what the agent updates with the DNS entries provided by the user, in the cloud-config) and the 'networkd' file from /run/resolvconf/interface and creating a unique list of 3 entries for the resolv.conf (this is what 'libc' does behind the scenes). I tried running this command with the 'libc' script set in debug mode and its basically what is placing 10.128.0.2 in my case at the top of the list. 

Anyways, digging deeper into the code, I also see the agent updating the 'dhclient.conf' file with the correct entries for DNS and also the '/etc/systemd/network' folder with a network file containing the correct DNS entries. So 2 questions:

1) If the dhclient.conf is being udapted - why is the agent not calling 'dhclient' to update the DNS (which in turn calls 'resolvconf' anyways)? I tried manually executing it in the VM and I got the correct data being fed into the resolv.conf.

2) Instead of updating the 'base' file, should the agent be updating the 'head' file in this scenario (where DNS servers are provided), which would guarantee the entries are added to the top of the file?

I am attaching the agent log for your reference, and also snippets of information about the files in question from the VM:

[bosh-agent-logs-2020-02-01.log](https://github.com/cloudfoundry/bosh-agent/files/5907275/bosh-agent-logs-2020-02-01.log)

```
nginx/7a6c11c1-57dc-4c57-ad7f-0305e9b2c168:/var/vcap/bosh/log# cat /run/resolvconf/interface/networkd 
nameserver 10.128.0.2
nameserver 10.128.1.11 10.128.2.12 10.129.0.13 10.128.0.2
search us-west-2.compute.internal

cat /run/resolvconf/resolv.conf 
nameserver 10.128.0.2
nameserver 10.128.1.11
nameserver 10.128.2.12
search us-west-2.compute.internal

nginx/7a6c11c1-57dc-4c57-ad7f-0305e9b2c168:/var/vcap/bosh/log# ls -l --full-time /run/resolvconf/resolv.conf 
-rw-r--r-- 1 root root 102 2021-02-01 21:24:38.208000000 +0000 /run/resolvconf/resolv.conf
```

Justin-W:
resolv.conf: ubuntu: Sync DNS nameserver changes to 'resolv.conf.d/head' (instead of 'resolv.conf.d/base'). Fixes #240.

NOTE: #240 does not require any changes for centos, because the resolv.conf is managed by Network Manager (and not updated by the bosh agent) on centos/rhel.

NOTE: When DNS servers are configured, the resolve.conf update code now updates 'resolv.conf.d/head' (instead of 'resolv.conf.d/base'). See #240 for details.

NOTE: When DNS servers are NOT configured, the existing behavior is unchanged.


ramonskie:
some more information about this

https://bugs.launchpad.net/ubuntu/+source/resolvconf/+bug/917896/comments/1
```
[quote]
Also significant for you may be the fact that since version 1.60 resolvconf includes dynamic nameserver information *before* information from the base file. In earlier versions the order was reversed. See Debian bug report #642222 (http://bugs.debian.org/642222) for the discussion that led up to this change.

Material that should precede dynamic and "base" material in resolv.conf should be placed in /etc/resolvconf/resolv.conf.d/head.
[/quote]
```

ramonskie (approved):

rkoster (changes requested):
Has this change been tested with bosh-dns? Because I'm worried that this change might break the bosh-dns resolv conf manager: https://github.com/cloudfoundry/bosh-dns-release/blob/master/src/bosh-dns/dns/manager/resolv_conf_manager.go#L73

If bosh-dns appends itself to the bottom of that file after the DNS servers from configured via cloud-config, bosh-dns will no longer work (it needs to be the first entry in resolv.conf).

rkoster:
bosh-dns will recurse to the dns servers in the resolv.conf if the DNS name is not a bosh DNS host. So in those cases it basically acts as a proxy. But for that to work it needs to be the first entry in the resolv.conf.

Justin-W:
FYI: The core issue seems to be that [this Ubuntu change](https://bugs.launchpad.net/ubuntu/+source/resolvconf/+bug/917896/comments/1) breaks DNS on AWS when DHCP Option Sets are enabled, because the agent's current behavior of writing to the 'resolv.conf.d/base' file now results in the Bosh DNS servers having lower precedence than the AWS name servers that get injected into the 'resolv.conf' due to AWS DHCP Option Sets.

The fix in this PR (as originally submitted) fixes the issue above, but creates a conflict between bosh-agent and bosh-dns, since both of them would attempt to own and modify the 'resolv.conf.d/head' file (and because the existing bosh-dns code doesn't work correctly if that file already exists and has content).

I will need to do some further investigation and testing about how to resolve the conflict between this PR and bosh-dns.

PR #278, opened 2022-04-07
Re-add support for Centos and RHEL stemcells

Justin-W:
This re-adds the previously-removed support for Centos stemcells (and also RHEL stemcells, since both types are handled identically by the agent).

NOTE: A PR to re-add https://github.com/cloudfoundry/bosh-linux-stemcell-builder support for Centos and RHEL stemcells will be submitted soon.


jpalermo (approved):
Looks good to me.

I'm guessing the `guide-rails.json` is for a CI pipeline you'll be running yourselves? It feels a bit out of place, but not a huge deal. Major concern would be a year or two go by, somebody figures it's unused and deletes it out from under you.

If you are interested in adding integration tests at some point, we'll probably want to rename some of the current CI tasks to avoid confusion. `test-integration` would be more clear as `test-integration-ubuntu` at this point.

Justin-W:
> I'm guessing the `guide-rails.json` is for a CI pipeline you'll be running yourselves?

Yes, that is the CI control file for our CI system.

beyhan (approved):

PR #279, opened 2022-04-11
Introduce error handling for nats disconnects and add logging handlers for nats connection issues

nouseforaname:
We are seeing intermittent unresponsive agents in following scenario:

- upgrade bosh-director
- right after upgrade is finished, start recreating a deployment

In some situations there seems to be a race condition in between

1. the old director going down
2. Agent on the VM retries sending heartbeets
3. new director comes up and get's deploy command issued. Sends ping to agent
4. ???? Currently unclear why at this point the agent has not yet reconnected to nats
5. director doesn't get a response and the deploy command fails.

the issue has been solved in CI by introducing a 3 minute wait after the upgrade 
of the director but we would like to get to the root cause of this. It seems that if 
the agent DID NOT crash and restart within BOSHs downtime, the connection is 
somehow in a bad state. Cunnie suggested it that the arp cache could be the 
reason the connection is not successfully re-established after the director comes 
back up.

Use nats client handler functions to add additional logging and an arp flush on 
connection issues.

In case the arp cache should not be issue at hand, the introduced routine can be 
changed to solve the actual problem

Tracker: **[ 181274434, #181759211]**

### DIFF
--- a/mbus/nats_handler.go
+++ b/mbus/nats_handler.go
@@ -118,7 +118,18 @@ func (h *natsHandler) Run(handlerFunc boshhandler.Func) error {
 
 	return nil
 }
+func asyncErrorHandler(natsErrorHandlingChan chan string, logger boshlog.Logger, p boshplatform.Platform, connInfo ConnectionInfo) {
+	logger.Debug(natsHandlerLogTag, "Received async nats log")
+	err := p.DeleteARPEntryWithIP(connInfo.IP)
+	if err != nil {
+		logger.Debug(natsHandlerLogTag, fmt.Sprintf("Delete ArpEntries for %v, Error: %v", connInfo.IP, err.Error()))
+	}
+	for value := range natsErrorHandlingChan {
+		time.Sleep(time.Second)
+		logger.Debug(natsHandlerLogTag, fmt.Sprintf("%v", value))
+	}
 
+}
 func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 	h.RegisterAdditionalFunc(handlerFunc)
 
@@ -126,6 +137,8 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 	if err != nil {
 		return bosherr.WrapError(err, "Getting connection info")
 	}
+	natsErrorHandlingChan := make(chan string)
+	go asyncErrorHandler(natsErrorHandlingChan, h.logger, h.platform, *connectionInfo)
 
 	h.logger.Info(h.logTag, "Attempting to connect to NATS")
 
@@ -142,6 +155,12 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 		nats.ReconnectWait(h.connectRetryInterval),
 		nats.ReconnectJitter(h.maxConnectRetryInterval, h.maxConnectRetryInterval),
 		nats.Secure(connectionInfo.TLSConfig),
+		nats.DisconnectHandler(func(c *nats.Conn) {
+			for c.IsReconnecting() {
+				natsErrorHandlingChan <- fmt.Sprintf("Waiting to reconnect to nats.. Connected: %v", c.IsConnected())
+			}
+			natsErrorHandlingChan <- "Reconnected to nats"
+		}),
 	}
 
 	connection, err := h.connector(connectionInfo.Addr, natsOptions...)
--- a/mbus/nats_handler.go
+++ b/mbus/nats_handler.go
@@ -33,7 +33,8 @@ const (
 	natsHandlerLogTag        = "NATS Handler"
 	natsMinRetryWait         = 2
 	natsConnectionMaxRetries = 10
-	natsMaxReconnectWait     = 120 * time.Second
+	//natsMaxReconnectWait should match the setting we have in BOSH for https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/agent_client.rb#L44. For now defaulting to 45 seconds. We could propagate this value in the settings.json with the other mbus settings.
+	natsMaxReconnectWait = 45 * time.Second
 )
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
@@ -136,7 +137,6 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 		nats.DisconnectErrHandler(func(c *nats.Conn, err error) {
 			h.logger.Debug(natsHandlerLogTag, "Nats disconnected with Error: %v", err.Error())
 			h.logger.Debug(natsHandlerLogTag, "Attempting to reconnect: %v", c.IsReconnecting())
-			h.arpClean()
 			for c.IsReconnecting() {
 				h.logger.Debug(natsHandlerLogTag, fmt.Sprintf("Waiting to reconnect to nats.. Connected: %v", c.IsConnected()))
 				time.Sleep(time.Second)
@@ -147,7 +147,6 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 		}),
 		nats.ClosedHandler(func(c *nats.Conn) {
 			h.logger.Debug(natsHandlerLogTag, "Connection Closed with: %v", c.LastError().Error())
-			h.arpClean()
 		}),
 		nats.ErrorHandler(func(c *nats.Conn, s *nats.Subscription, err error) {
 			h.logger.Debug(natsHandlerLogTag, err.Error())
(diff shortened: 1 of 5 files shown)


mbus/nats_handler.go
```diff
@@ -118,14 +118,27 @@ func (h *natsHandler) Run(handlerFunc boshhandler.Func) error {
 
 	return nil
 }
+func asyncErrorHandler(natsErrorHandlingChan chan string, logger boshlog.Logger, p boshplatform.Platform, connInfo ConnectionInfo) {
+	logger.Debug(natsHandlerLogTag, "Received async nats log")
+	err := p.DeleteARPEntryWithIP(connInfo.IP)
+	if err != nil {
+		logger.Debug(natsHandlerLogTag, fmt.Sprintf("Delete ArpEntries for %v, Error: %v", connInfo.IP, err.Error()))
+	}
```
ctlong, line 126: Shouldn't the above be in the loop below?

We think the way `asyncErrorHandler` is currently structured, the ARP flush will occur when the goroutine is first started, rather than anytime an error is received in the `natsErrorHandlingChan`.

mbus/nats_handler.go
```diff
@@ -142,6 +155,12 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 		nats.ReconnectWait(h.connectRetryInterval),
 		nats.ReconnectJitter(h.maxConnectRetryInterval, h.maxConnectRetryInterval),
 		nats.Secure(connectionInfo.TLSConfig),
+		nats.DisconnectHandler(func(c *nats.Conn) {
+			for c.IsReconnecting() {
+				natsErrorHandlingChan <- fmt.Sprintf("Waiting to reconnect to nats.. Connected: %v", c.IsConnected())
```
ctlong, line 160: Could we just inline the `asyncErrorHandler` code here and eliminate all the channel / goroutine stuff?

rkoster:
Related discussion here: https://github.com/cloudfoundry/bosh/discussions/2372

beyhan:
Should we first add logging as suggested in the title of this pr to understand the issue and afterwards try to fix it? It could be that adding the cleanup of the ARP cache can hide the real issue.

nouseforaname:
> 

Not sure if it really could hide other Issues. We know that  on a recereate the Mac Address on the Director will change and consequent comms will fail until the clients will renew that IP-Mac mapping. I don't think it will hide any other Issues but it will definitely solve one we're aware of.

lnguyen (approved):
LGTM overall, we can we rename PR, this really isn't adding logging to nats disconnect but rather clean up arp on nats disconnect

mbus/nats_handler.go
```diff
@@ -130,17 +130,35 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
...
+			h.logger.Debug(natsHandlerLogTag, err.Error())
+		}),
+		nats.CustomReconnectDelay(func(attempts int) time.Duration {
+			reconnectWait := time.Duration(time.Duration(math.Pow(natsMinRetryWait, float64(attempts))) * time.Second)
+			h.logger.Debug(natsHandlerLogTag, fmt.Sprintf("Increased reconnect to: %v", reconnectWait))
+			return reconnectWait
+		}),
 		nats.MaxReconnects(natsConnectionMaxRetries),
-		nats.ReconnectWait(h.connectRetryInterval),
-		nats.ReconnectJitter(h.maxConnectRetryInterval, h.maxConnectRetryInterval),
```
beyhan, line 143: We don't define any `ReconnectJitter` any more but will use the default one https://github.com/nats-io/nats.go/blob/1714729895dfe3f5aa0cfee3984b89bebcfce23c/nats.go#L57. Is this wanted?

mbus/nats_handler.go
```diff
@@ -130,17 +130,35 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
...
-		nats.Timeout(natsConnectionTimeout),
+		nats.DisconnectHandler(func(c *nats.Conn) {
+			for c.IsReconnecting() {
+				h.arpClean()
+				h.logger.Debug(natsHandlerLogTag, fmt.Sprintf("Waiting to reconnect to nats.. Connected: %v", c.IsConnected()))
+				time.Sleep(time.Second)
+			}
+			h.logger.Debug(natsHandlerLogTag, "Reconnected to nats")
+		}),
+		nats.ClosedHandler(func(c *nats.Conn) {
```
beyhan, line 144: My understanding is that `ClosedHandler` is called when the NATs client failed to create a connection or to reconnect. Is the reconnection use case possible here?

mbus/nats_handler.go
```diff
@@ -130,17 +130,35 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
...
-		err = h.platform.DeleteARPEntryWithIP(connectionInfo.IP)
-		if err != nil {
-			h.logger.Error(h.logTag, "Cleaning ip-mac address cache for: %s", connectionInfo.IP)
-		}
+		h.arpClean()
 	}
-
 	var natsOptions = []nats.Option{
-		nats.Timeout(natsConnectionTimeout),
+		nats.DisconnectHandler(func(c *nats.Conn) {
```
beyhan, line 136: Why we use the deprecated `DisconnectHandler` and not the replacement `DisconnectErrHandler? See https://github.com/nats-io/nats.go/pull/504/files

beyhan (changes requested):
In general I will appreciate to have a list with all settings explained in the description which this pr changes. At the moment the description of this pr is not complete.

mbus/nats_handler.go
nouseforaname: That was replaced by adding the exponential backoff. Looking through the old commit messages the Idea was to introduce some variation to lessen the load on the nats server in case of restart.

mbus/nats_handler.go
```diff
@@ -28,12 +29,10 @@ import (
 )
 
 const (
-	responseMaxLength           = 1024 * 1024
-	natsHandlerLogTag           = "NATS Handler"
-	natsConnectionTimeout       = 5 * time.Minute
```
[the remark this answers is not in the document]
nouseforaname: This 5 minute timeout was introduced to avoid agents crashing when switching from bosh retry to nats retry. 

https://github.com/cloudfoundry/bosh-agent/commit/9b88d9fffeab52221aa0bbb5c633c9a9876cbf91

The Issues with the `agent timed out` on deployments after a bosh create-env appeared right after.

mbus/nats_handler.go
nouseforaname: Nice natch. I haven't seen the deprecation notice. 
Thanks.

mbus/nats_handler.go
beyhan: @nouseforaname any comments on this?

nouseforaname: yes, you have a valid point there. I updated the code to only log the last error here. I updated the code to match the nats example from:
https://github.com/nats-io/nats.go#clustered-usage

nouseforaname:
I updated the maxReconnectWait to match the default 45 seconds from 
https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/agent_client.rb#L44

This is to avoid having a reconnectWait longer than the director would wait for the agent to respond in the following sequence of events:

t0s: Current wait: 32s. agent reconnect fails. wait-> 64s
t10s: bosh is fully up, receives deploy command, pings agent
t55s: bosh gives up waiting for agent
t64s: agent tries to reconnect but it's too late for the deployment.

mbus/nats_handler.go
```diff
@@ -115,32 +120,47 @@ func (h *natsHandler) Run(handlerFunc boshhandler.Func) error {
...
-		err = h.platform.DeleteARPEntryWithIP(connectionInfo.IP)
-		if err != nil {
-			h.logger.Error(h.logTag, "Cleaning ip-mac address cache for: %s", connectionInfo.IP)
-		}
+		h.arpClean()
 	}
-
 	var natsOptions = []nats.Option{
-		nats.Timeout(natsConnectionTimeout),
+		nats.DisconnectErrHandler(func(c *nats.Conn, err error) {
```
beyhan, line 137: * We don't clean the ARP cache any more on reconnect. This get lost somehow. 
* Can we block in a *handler with the `for` loop? I don't see this pattern to be used in their samples.

beyhan (changes requested):

mbus/nats_handler.go
nouseforaname: I've been looking into the nats code. I think it's only blocking additional calls of error handler but I'm not quite sure yet.
The function ends up in a `sync.Cond` and looking at the logs the agent produces with the change it doesn't seem it's blocking other actions in the agent itself

nouseforaname: As far as the repeated flush goes, I'm not sure that is necessary more than once as the first flush already removes the old Mac from the cache and success we cannot reconnect in but sure it needs to be flushed again.

beyhan: Ok, this means you are still looking into this. Thanks for the update!

beyhan (approved):

ctlong (approved):

Issue #289, opened 2022-09-08
connection flooding from bosh-agents whilst director is unavailable


guzzisti:
# observed behaviour
When a director is down for maintenance, every VM with a bosh agent starts to flood the director with connection requests to port 4222 after a while (~ 5 minutes in our environment). We see up to a several hundred connection attemps **per second** (sic!) from every vm managed by bosh.
After several minutes of packet storm it will calm down, just to start again after a short time of normal operation.
The graph shows the number of connection attempts by a single vm per second:
![image](https://user-images.githubusercontent.com/10488791/189072295-e9d8125e-8b25-408e-a60c-c0f6c669ed54.png)

This causes severe load on the network components, depending on how large your bosh environment is.

The bosh agent log shows the section:
```
2022-09-07_12:27:56.70086 [NATS Handler] 2022/09/07 12:27:56 DEBUG - Increased reconnect to: 2s
2022-09-07_12:28:00.70185 [NATS Handler] 2022/09/07 12:28:00 DEBUG - Increased reconnect to: 4s
2022-09-07_12:28:06.70327 [NATS Handler] 2022/09/07 12:28:06 DEBUG - Increased reconnect to: 8s
2022-09-07_12:28:16.70370 [NATS Handler] 2022/09/07 12:28:16 DEBUG - Increased reconnect to: 10s
2022-09-07_12:28:28.70543 [NATS Handler] 2022/09/07 12:28:28 DEBUG - Increased reconnect to: 10s
2022-09-07_12:28:40.71067 [NATS Handler] 2022/09/07 12:28:40 DEBUG - Increased reconnect to: 10s
2022-09-07_12:28:52.71286 [NATS Handler] 2022/09/07 12:28:52 DEBUG - Increased reconnect to: 10s
2022-09-07_12:29:02.71468 [NATS Handler] 2022/09/07 12:29:02 DEBUG - Increased reconnect to: 10s
2022-09-07_12:29:12.71673 [NATS Handler] 2022/09/07 12:29:12 DEBUG - Increased reconnect to: 10s
2022-09-07_12:29:22.71787 [NATS Handler] 2022/09/07 12:29:22 DEBUG - Increased reconnect to: 10s
2022-09-07_12:29:32.72396 [NATS Handler] 2022/09/07 12:29:32 DEBUG - Increased reconnect to: 10s
[...]

PR #291, opened 2022-09-08
fix overflow bug in nats_handler.go / nats.CustomReconnectDelay()

daniel-hoefer:
fix for https://github.com/cloudfoundry/bosh-agent/issues/289
- avoid arithmetic overflow with math.Pow and subsequent conversion to int64 (by limiting exponent to a safe value)
- add sanity check for reconnect delay to safeguard against negative values or zero
- add support for delay in non-integer seconds (by fixing how time.Duration() is used with time.Second)
- use more consistent naming of constants and variables

### DIFF
--- a/mbus/nats_handler.go
+++ b/mbus/nats_handler.go
@@ -154,13 +154,13 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
 			h.logger.Debug(natsHandlerLogTag, err.Error())
 		}),
 		nats.CustomReconnectDelay(func(attempts int) time.Duration {
-		    reconnectSeconds := natsMinReconnectSeconds * math.Pow(2.0, math.Min(20.0, float64(attempts - 1)))
-		    if (reconnectSeconds > natsMaxReconnectSeconds || reconnectSeconds <= 0.0) {
-		        reconnectSeconds = natsMaxReconnectSeconds
-		    }
-		    reconnectDelay := time.Duration(reconnectSeconds * float64(time.Second))
-		    h.logger.Debug(natsHandlerLogTag, "Increased reconnect to: %v", reconnectDelay)
-		    return reconnectDelay
+			reconnectSeconds := natsMinReconnectSeconds * math.Pow(2.0, math.Min(20.0, float64(attempts - 1)))
+			if (reconnectSeconds > natsMaxReconnectSeconds || reconnectSeconds <= 0.0) {
+				reconnectSeconds = natsMaxReconnectSeconds
+			}
+			reconnectDelay := time.Duration(reconnectSeconds * float64(time.Second))
+			h.logger.Debug(natsHandlerLogTag, "Increased reconnect to: %v", reconnectDelay)
+			return reconnectDelay
 		}),
 		nats.MaxReconnects(-1),
 		nats.Secure(connectionInfo.TLSConfig),
--- a/mbus/nats_handler.go
+++ b/mbus/nats_handler.go
@@ -31,9 +31,9 @@ import (
 const (
 	responseMaxLength = 1024 * 1024
 	natsHandlerLogTag = "NATS Handler"
-	natsMinRetryWait  = 2
-	//natsMaxReconnectWait should be lower than the setting we have in BOSH for https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/agent_client.rb#L44.
-	natsMaxReconnectWait = 10 * time.Second
+	natsMinReconnectSeconds = 2.0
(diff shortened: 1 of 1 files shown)


daniel-hoefer:
Sorry, just noticed that I've messed up the indentation inside `nats.CustomReconnectDelay()` due to the configuration of the IDE, using 4 spaces instead of tabs.

Regarding the exponential backoff: does it really make sense to use it when starting at a reconnect delay of 2s, but capping it at 10s? (i.e. the delay sequence is 2s, 4s, 8s, 10s, 10s, ...)

Maybe the exponential backoff made more sense before the cap (`natsMaxReconnectSeconds`) was introduced? I think line 157 could be simplified to `reconnectSeconds := natsMinReconnectSeconds * float64(attempts)`, only slightly changing the sequence to (2s, 4s, 6s, 8s, 10s, 10s, ...), but making the line easier to comprehend.

mbus/nats_handler.go
```diff
@@ -154,13 +154,13 @@ func (h *natsHandler) Start(handlerFunc boshhandler.Func) error {
...
 		}),
 		nats.CustomReconnectDelay(func(attempts int) time.Duration {
-			exponentialReconnectWait := time.Duration(math.Pow(natsMinRetryWait, float64(attempts))) * time.Second
-			if natsMaxReconnectWait > exponentialReconnectWait {
-				h.logger.Debug(natsHandlerLogTag, "Increased reconnect to: %v", exponentialReconnectWait)
-				return exponentialReconnectWait
-			}
-			h.logger.Debug(natsHandlerLogTag, "Increased reconnect to: %v", natsMaxReconnectWait)
-			return natsMaxReconnectWait
+		    reconnectSeconds := natsMinReconnectSeconds * math.Pow(2.0, math.Min(20.0, float64(attempts - 1)))
```
danielfor, line 157: Is it possible to remove the magic numbers? I propose replacing 
`math.Pow(2.0, `
with
`math.Pow(natsMinReconnectSeconds,`
And `20.0` with a new constant.

danielfor (changes requested):

jpalermo:
Do we understand why this is overflowing? If it's waiting 2^30 seconds, that's like 30 years, so I'm guessing something is either not waiting for the duration, or something else is going on here?

danielfor:
@jpalermo I think it's because we return `natsMaxReconnectWait` which is 10 seconds, if the `exponentialReconnectWait`  is more than the same 10 seconds, to get to 30 iterations we would need to wait only 300 seconds, and at that point we will be overflowing the datatype.

mbus/nats_handler.go
daniel-hoefer: `natsMinReconnectSeconds` needs to stay outside of `math.Pow()` to keep the behavior at least somewhat predictable. E.g. if someone changes `natsMinReconnectSeconds` to 10.0, by your proposal the delay sequence (disregarding `natsMaxReconnectSeconds`) with each attempt would be (1s, 10s, 100s, 1000s, ...) and an overflow would already occur when the exponent, i.e. `attempts - 1`, equals 10.

Also, I'd expect exponential backoff to use base 2, i.e. doubling the delay with each retry (instead of e.g. tripling it or using any factor other than 2). Doubling means that the first argument to `math.Pow()` needs to be equal to 2.0. This is another reason why `natsMinReconnectSeconds` should be multiplicated with `math.Pow(...)`, instead of being passed as the first argument to it.

I agree that keeping 20.0 as a magic number in the `math.Pow(...)` expression is not ideal. But coming up with a good name for the constant and justifying the value is quite hard.

Since understanding the exponential backoff formula and predicting when overflow occurs is so complicated, I'm really favoring a much simpler backoff mechanism, e.g. `reconnectSeconds := natsMinReconnectSeconds * float64(attempts)`.

I'll push a new commit for the whitespace fix and another for the simpler backoff mechanism. Let me know if you're fine with this.

daniel-hoefer:
Sorry for "requested review from danielfor and removed request for jpalermo and lnguyen" ( https://github.com/cloudfoundry/bosh-agent/pull/291#event-7366062937 ) - that was unintentional. I'm quite new to GitHub, and still figuring out the details.

danielfor (approved):
LGTM

lnguyen (changes requested):
LGTM can you fix linting issues

daniel-hoefer:
> LGTM can you fix linting issues

Sorry, the linter errors seem to be completely unrelated to my changes. I don't even know where to start fixing them. Someone more experienced with Go and with this project needs to do this.

jpalermo:
Yeah, I don't know what is up with the linting errors, but the changes look good. Thanks!

PR #293, opened 2022-10-19
Fix test-integration-windows task.

danielfor:
Path searching changed with go 1.19 in a way that having blobstore executables in our current path (`c:\bosh`) will interfere with the test correctly locating the binaries in the folder `c:\var\vcap\bosh\bin` folder.

We don't believe these binaries are used any longer.

Signed-off-by: Daniel Felipe Ochoa <danielfelipo@vmware.com>


ragaskar:
LGTM!

PR #294, opened 2022-10-19
Fix for test-integration-windows

danielfor:
Removing tar.exe to avoid conflicts with binaries in the current folder and binaries in the path. See previous commit
(626ed6b4dac044cb3c5b6c896d482c0ad70908e8) for more info

Signed-off-by: Rajan Agaskar <ragaskar@vmware.com>


ragaskar:
LGTM. 

This may make the Vagrantfile in the project no longer work as intended, but it doesn't look like that's been maintained. We should follow up and remove it if it is obsolete and/or fix it if it is still in use.

PR #295, opened 2022-10-26
Remove nosuid flag from /opt and /var/opt

jpalermo:
The work in [#254] and [#256] to mount `/opt` and `/var/opt` on the ephemeral disk accidentally added the nosuid flag to the mount.

We had a report that this is causing problems for bosh releases that use these folders.


cunnie (approved):
LGTM

lnguyen (approved):
LGTM

PR #297, opened 2022-11-17
Preserve extended attributes when migrating disk data

speer:
The following commit tried to fix the preservation of extended attributes when migrating disk data, however is not complete, as also the `--xattrs-include=*.*` parameter is required to achieve for example the preservation of capabilities: https://github.com/cloudfoundry/bosh-agent/commit/c3f558b4d705c0d78b624bfc2b378901d338600b
    
With this we try to finally fix the preservation of all extended attributes, such as capabilities for example.
    
Preservation of capabilities are for importance in PKS (Tanzu Kubernetes Grid Integrated Edition / TKGI), as Docker images are stored on persistent disks.
Whenever a Docker image includes a binary with some capabilities (for example nginx-ingress-controller), these capabilities, without this change, get lost after disk migrations, which prevents the successful starting of containers after the disk migration.
    
So far in TKGI all docker images need to be manually wiped from disk, after disk migration and re-pulled, so that the capabilities get instanciated.
    
Behaviour comparison of both commands (current vs. proposed):
```
$ getcap original/nginx-ingress-controller
original/nginx-ingress-controller = cap_net_bind_service+ep

$ (tar -C original --xattrs -cf - .) | (tar -C copy --xattrs -xpf -)
$ getcap copy/nginx-ingress-controller
$

$ (tar -C original --xattrs --xattrs-include=*.* -cf - .) | (tar -C copy --xattrs --xattrs-include=*.* -xpf -)
$ getcap copy/nginx-ingress-controller
copy/nginx-ingress-controller = cap_net_bind_service+ep
```

For further reference:
* https://www.gnu.org/software/tar/manual/html_node/Extended-File-Attributes.html
* VMware Bug: BOSH-180

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -1380,7 +1380,8 @@ func (p linux) MigratePersistentDisk(fromMountPoint, toMountPoint string) error 
 
 	// Golang does not implement a file copy that would allow us to preserve dates...
 	// So we have to shell out to tar to perform the copy instead of delegating to the FileSystem
-	tarCopy := fmt.Sprintf("(tar -C %s --xattrs -cf - .) | (tar -C %s --xattrs -xpf -)", fromMountPoint, toMountPoint)
+	// The --xattrs and --xattrs-include=*.* flags ensure that all extended attributes (ex. capabilities) are preserved
+	tarCopy := fmt.Sprintf("(tar -C %s --xattrs --xattrs-include=*.* -cf - .) | (tar -C %s --xattrs --xattrs-include=*.* -xpf -)", fromMountPoint, toMountPoint)
 	_, _, _, err = p.cmdRunner.RunCommand("sh", "-c", tarCopy)
 	if err != nil {
 		return bosherr.WrapError(err, "Copying files from old disk to new disk")
--- a/platform/linux_platform_test.go
+++ b/platform/linux_platform_test.go
@@ -3674,7 +3674,7 @@ sam:fakeanotheruser`)
 			Expect(mounter.RemountAsReadonlyArgsForCall(0)).To(Equal("/from/path"))
 
 			Expect(len(cmdRunner.RunCommands)).To(Equal(1))
-			Expect(cmdRunner.RunCommands[0]).To(Equal([]string{"sh", "-c", "(tar -C /from/path --xattrs -cf - .) | (tar -C /to/path --xattrs -xpf -)"}))
+			Expect(cmdRunner.RunCommands[0]).To(Equal([]string{"sh", "-c", "(tar -C /from/path --xattrs --xattrs-include=*.* -cf - .) | (tar -C /to/path --xattrs --xattrs-include=*.* -xpf -)"}))
 
 			Expect(mounter.UnmountCallCount()).To(Equal(1))
 			Expect(mounter.UnmountArgsForCall(0)).To(Equal("/from/path"))
@@ -3734,7 +3734,7 @@ from-device-path  dm-0 NETAPP  ,LUN C-Mode
 				Expect(mounter.RemountAsReadonlyArgsForCall(0)).To(Equal("/from/path"))
 
 				Expect(len(cmdRunner.RunCommands)).To(Equal(3))
-				Expect(cmdRunner.RunCommands[0]).To(Equal([]string{"sh", "-c", "(tar -C /from/path --xattrs -cf - .) | (tar -C /to/path --xattrs -xpf -)"}))
+				Expect(cmdRunner.RunCommands[0]).To(Equal([]string{"sh", "-c", "(tar -C /from/path --xattrs --xattrs-include=*.* -cf - .) | (tar -C /to/path --xattrs --xattrs-include=*.* -xpf -)"}))
 
 				Expect(mounter.UnmountCallCount()).To(Equal(1))
 				Expect(mounter.UnmountArgsForCall(0)).To(Equal("/from/path"))


rkoster:
@speer could you add a test to demonstrate the case this code is fixing?

speer:
> @speer could you add a test to demonstrate the case this code is fixing?

What do you mean exactly by test? We have PKS/TKGI and whenever disk migrations happen (ex. increase persistent disk size via Opsman), binaries within Docker images (ex. nginx-ingress-controller) loose the capabilities they had before.
As stated in the commit message `--xattrs-include=*.*` preserves capabilities and should solve this problem.

platform/linux_platform.go
```diff
@@ -1380,7 +1380,7 @@ func (p linux) MigratePersistentDisk(fromMountPoint, toMountPoint string) error
 
 	// Golang does not implement a file copy that would allow us to preserve dates...
 	// So we have to shell out to tar to perform the copy instead of delegating to the FileSystem
-	tarCopy := fmt.Sprintf("(tar -C %s --xattrs -cf - .) | (tar -C %s --xattrs -xpf -)", fromMountPoint, toMountPoint)
+	tarCopy := fmt.Sprintf("(tar -C %s --xattrs --xattrs-include=*.* -cf - .) | (tar -C %s --xattrs --xattrs-include=*.* -xpf -)", fromMountPoint, toMountPoint)
```
danielfor, line 1383: When you specify `*.*` you will be including the extended attributes of all the files that contain a dot in the name. I think this will not include files without a dot, for example: `Makefile` or `LICENSE`. Is the intention to leave this files out? If not, it might be better to use `--xattrs-include=*`

speer: `--xattrs-include` is not referring to filenames, as I understood [this](https://www.gnu.org/software/tar/manual/html_node/Extended-File-Attributes.html).
```
Here, the pattern is a globbing pattern. For example, the following command:
$ tar --xattrs --xattrs-exclude='user.*' -c a.tar .
will include in the archive ‘a.tar’ all attributes, except those from the ‘user’ namespace.

...

Attribute names are strings prefixed by a namespace name and a dot. Currently, four namespaces exist: ‘user’, ‘trusted’, ‘security’ and ‘system’.
```

This is why I think `*.*` should be correct. I even tested it with a file named `nginx-ingress-controller` which doesn't include a dot.

danielfor: thanks for the clarification.

danielfor, line 1383: This change breaks the following unit tests: 

`bosh-agent/platform/linux_platform_test.go:3704`
`bosh-agent/platform/linux_platform_test.go:3669`

Could you please fix them according to the new behavior and, as Ruben suggested, add a new one describing the expected behavior? For more info look [here](https://github.com/cloudfoundry/bosh-agent/blob/main/docs/dev_setup.md#running-tests).

speer: Thank's for spotting and explaining.
Fixed the tests:
```
Ginkgo ran 51 suites in 2m42.468308791s
Test Suite Passed

SUCCESS
```

Regarding the creation of a new test, I lack golang skills and deeper knowledge in this project, as I am not sure how to achieve this.
I would assume it would require an extension of the [FakeFileSystem](https://github.com/cloudfoundry/bosh-utils/blob/master/system/fakes/fake_file_system.go), as that one is not aware about extended attributes in general at the moment.
Is there already somewhere a test for the existing `--xattrs` flag, which I could somehow re-use and do things similar?

danielfor: @speer Looking at the code I see that creating a unit test for this might be a little hard because of the fake objects, the other route that could be taken is to create an integration test specific for linux, These kind of tests are designed to run on the target os (linux) , that could make things easier.
There is a _similar_ test that can be partially re used: `integration/ephemeral_disk_test.go`. 
@klakin-pivotal do you have any opinions on the testing (or not testing) of this feature?

klakin-pivotal: Given that the only tests I see for persistent disk migration are the tests in `migrate_disk_test.go` and `linux_platform_test.go` which use the `FakeFileSystem` class, it seems acceptable to me to not write an integration test for this and just update the `cmdRunner.RunCommands` expectations in the two tests that are failing because of the change to the `tar` invocation.

I had assumed that there was an integration test that tested Persistent Disk Migration, but now that I know that there's not one, I'm okay with just updating the existing tests.

I certainly wouldn't say "No" to a new integration test, and I agree that looking at `integration/ephemeral_disk_test.go` would be a reasonable way to begin to get an idea of how to create an integration test for this bug.

speer: > Given that the only tests I see for persistent disk migration are the tests in migrate_disk_test.go and linux_platform_test.go which use the FakeFileSystem class, it seems acceptable to me to not write an integration test for this and just update the cmdRunner.RunCommands expectations in the two tests that are failing because of the change to the tar invocation.

The two tests have been updated.
So would it be ok, if the new "persistent_disk" integration test would be created outside of this PR? I definitively lack knowledge in the project to be able to create such just by myself in reasonable time.

danielfor: Yeah, I think it can be done later. Thanks.

danielfor (approved):
LGTM

rkoster:
Thanks! @speer

klakin-pivotal:
Internal VMWare FYI: We failed to get this merged into the branch of the Agent that's used in our Xenial stemcells, disappointing some of our Corporate customers. So, we cherry-picked this commit and pushed it to the `2.268.x`. Here's the commit, just in case one of us at VMware comes looking one day 1f3763259a340d2bc09abddfea3bc19bb8ad8730

PR #301, opened 2023-02-09
Reorder code to boradcast address earlier

Lafunamor:
This reorder should help prevent cases in which the firewall rule for NATS cannot be setup due to missing connectivity
Resolves:
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/280

### DIFF
--- a/platform/net/ubuntu_net_manager.go
+++ b/platform/net/ubuntu_net_manager.go
@@ -185,12 +185,12 @@ func (net UbuntuNetManager) SetupNetworking(networks boshsettings.Networks, mbus
 		return bosherr.WrapError(err, "Validating dns configuration")
 	}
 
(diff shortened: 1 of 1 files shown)


jpalermo:
So it's [this](https://github.com/cloudfoundry/bosh-agent/blob/c8d9dbf857d9aab53889194fff4f4db8a8c2e639/platform/net/firewall_provider_linux.go#L66) lookup that potentially fails because the broadcast hasn't happened yet?

Seems correct to do it in the new order. Thanks!

jpalermo (approved):

jpalermo:
When you get a chance @Lafunamor, can you sign the CLA, or if you've already done so, submit a ticket to get help (sometimes it seems like the bot loses people)

Lafunamor:
@jpalermo I'm working on getting our companies CLA re-signed. Based on the feedback I got from the LF support there were some internal changes and the existing one is no longer recognized by the bot.

rkoster (approved):

PR #304, opened 2023-03-15
Handles "system" & "all" types in fetch log action

klakin-pivotal:
This commit updates the Bosh Agent to support the "system" and "all" log types that are now (or will soon be) supported by the Bosh CLI. This work is backwards-compatible with versions of the Bosh CLI that do not support these new log types.

The "system" log type fetches the logs from "/var/log". System logs are placed into a "/var/log" directory inside the tarball. This is done to generally avoid confusion as to the source of these logs, but especially when the "all" log type is requested. Do note that if the "all" log type is requested, and the VM has a job named "var" deployed, then the system logs will be placed into a "log" subdirectory in the "var" job's log directory. For now, this seems to be a minor and unlikely problem, so avoiding this doesn't seem to be worth the hassle.

The "all" log type is a pseudo-type constructed by the Bosh CLI, which fetches all currently-known log types; "job" (the historical (and current) default when no log type is specified), "agent", and "system".

NOTE: Requests for the "system" log type will be sliently ignored when
      the Agent is running on an OS that is not Linux. Currently, only
      Windows and Linux are supported OSs. If we ever support another
      Unix-alike OS (or figure out how to do something sensible and
      useful on Windows), then the code that handles the "system" log
      type will need to be updated.

NOTE: The fetch log action can handle combinations of log types that
      are not supported by the Bosh CLI. This should generally be
      harmless, and probably makes it easier to extend in the future.
      We also do no deduplication of the log types passed to the action.
      A malicious user could take advantage of this to cause the Agent
      to consume large amounts of RAM or disk space (whether in the
      blobstore, or on the Agent's disk), but if a malicious user has
      the credentials required to talk to the Agent, they can already
      get up to significant mischief.

Signed-off-by: Brian Cunnie <bcunnie@vmware.com

### DIFF
--- a/agent/action/fetch_logs.go
+++ b/agent/action/fetch_logs.go
@@ -36,8 +36,8 @@ func (a FetchLogsAction) IsLoggable() bool {
 	return true
 }
 
-func (a FetchLogsAction) Run(logType string, filters []string) (value map[string]string, err error) {
-	tarball, err := a.logsTarProvider.Get(logType, filters)
+func (a FetchLogsAction) Run(logTypes string, filters []string) (value map[string]string, err error) {
+	tarball, err := a.logsTarProvider.Get(logTypes, filters)
 	if err != nil {
 		return
 	}
--- a/agent/action/fetch_logs_with_signed_url_test.go
+++ b/agent/action/fetch_logs_with_signed_url_test.go
@@ -3,10 +3,9 @@ package action_test
 import (
 	"errors"
 
+	fakelogstarprovider "github.com/cloudfoundry/bosh-agent/agent/logstarprovider/logstarproviderfakes"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
-	fakelogstarprovider "github.com/cloudfoundry/bosh-agent/agent/logstarprovider/logstarproviderfakes"
 
 	boshassert "github.com/cloudfoundry/bosh-utils/assert"
 	boshcrypto "github.com/cloudfoundry/bosh-utils/crypto"
--- a/agent/logstarprovider/logs_tar_provider.go
+++ b/agent/logstarprovider/logs_tar_provider.go
@@ -1,10 +1,12 @@
 package logstarprovider
 
 import (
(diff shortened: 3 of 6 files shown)


klakin-pivotal:
NOTE: This PR is related to, but does not require this `bosh-cli` PR: https://github.com/cloudfoundry/bosh-cli/pull/614

klakin-pivotal:
@selzoc I'm tagging you because you recently refactored part of the code that I modified, so you may (or may not) wish to review that part of this change to ensure that I didn't break your change.

lnguyen (approved):
LGTM

klakin-pivotal:
> That struct still has LogType, singular. Is this intentional?

Yeah, I was being lazy and didn't want to track down the users of that field. Would you prefer that I do and rename the field?

selzoc:
> Yeah, I was being lazy and didn't want to track down the users of that field. Would you prefer that I do and rename the field?

Maybe as a fast-follow to these commits, I don't think we need to block any merging on it :)

klakin-pivotal:
@selzoc are you willing to approve these changes, with the caveat that I change the spelling of that variable that you mentioned in `Fetch LogsWithSignedUrlAction:`? (I think we're supposed to have approval from two reviewers before merging a PR, and I note that the identical changes in the PR for the other branch was merged in this morning. It'd be nice to be able to merge this in and get started on that rename and also CLI documentation.)

selzoc (approved):

selzoc:
Done!

klakin-pivotal:
@selzoc Because `LogType` seems to refer to a JSON key of the same name in these two spots:

https://github.com/cloudfoundry/bosh-agent/blob/2f41166593e281eacc99dea18b17b98dc246555c/agent/action/bundle_logs.go#L20
https://github.com/cloudfoundry/bosh-agent/blob/2f41166593e281eacc99dea18b17b98dc246555c/agent/action/fetch_logs_with_signed_url.go#L15

I'm going to leave the variable name as-is. I'd rather the variable name be a little incorrect than it not match its data source.

PR #306, opened 2023-03-17
Correctly test on non-Linux

klakin-pivotal:
When not running on Linux, we silently ignore requests for the "system" log type. We didn't account for this in the relevant test. This commit fixes that oversight.


klakin-pivotal:
This is a trivial and obvious correction (and I tested the `main` branch correction in CI), so I'm skipping the review and just pushing the "merge" button.

PR #308, opened 2023-04-12
remove the limit of characters for the device id

ramonskie:
as disk ids can be longer than 20 characters it seems strange that we put a limit on this. as that means that the disk-id glob search will never be found if it cuts of everything after 20 characters


cunnie (approved):
LGTM.

I'm also mystified why we truncated it.

klakin-pivotal:
Yeah, there's no historical justification for the choice to only use the first twenty characters, and I can't think of a reason to do things that way, so let's see what happens when we stop doing that, I guess.

klakin-pivotal (approved):

PR #309, opened 2023-05-04
correct the diskID for device path resolution

fmoehler:
For ali and aws the device path is not correctly resolved, due to some differences in the ID/Name of the disk between /dev/disk/by-id/ and the ID/Name of the disk in the IAAS provider

### DIFF
--- a/infrastructure/devicepathresolver/id_device_path_resolver_test.go
+++ b/infrastructure/devicepathresolver/id_device_path_resolver_test.go
@@ -28,7 +28,7 @@ var _ = Describe("IDDevicePathResolver", func() {
 		udev = fakeudev.NewFakeUdevDevice()
 		fs = fakesys.NewFakeFileSystem()
 		diskSettings = boshsettings.DiskSettings{
-			ID: "fake-disk-id-include-longname",
+			ID: "some-fake-disk-id-include-longname",
 		}
 	})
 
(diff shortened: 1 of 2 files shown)


infrastructure/devicepathresolver/id_device_path_resolver_test.go
```diff
@@ -57,7 +57,7 @@ var _ = Describe("IDDevicePathResolver", func() {
 				err = fs.Symlink("/dev/intermediate/fake-device-path", "/dev/disk/by-id/virtio-fake-disk-id-include-longname")
 				Expect(err).ToNot(HaveOccurred())
 
-				fs.SetGlob("/dev/disk/by-id/*fake-disk-id-include-longname", []string{"/dev/disk/by-id/virtio-fake-disk-id-include-longname"})
```
klakin-pivotal, line 60: I might just misunderstand the PR... but why did you remove `fake-` from here and elsewhere? Does that have anything to do with the change?

fmoehler: Maybe also this can be changed to be clearer. The different cloud providers usually have a disk ID starting with some characters, then a hyphen character and then continuing with a random string. (e.g. AWS: vol-123456789ABCD). However the symlink in by-id does not have the hyphen character (for AWS (e.g. nvme-Amazon_Elastic_Block_Store_vol123456789ABCD). For Ali, everything before and including the hyphen character is missing in by-id (e.g. Ali DiskID: d-123456789ABCD; by-id: virtio-123456789ABCD). This means the device path resolution always fails atm because the Glob looks for "*vol-123456789ABCD" (for AWS) or "*d-123456789ABCD" (for Ali), but cannot find it in by-id. Therefore, we remove everything before the hyphen to make the Glob work again. In this case this means removing the "fake-" at the beginning of the mocked fake disk id. Do you think the mocked Disk ID should be changed to be clearer? Something like "fake_disk-some_random_identifier"?

max-soe:
@klakin-pivotal @nouseforaname @cunnie,
This is an important bugfix for us. Can you please review this PR?

nouseforaname:
Hi,

can you please provide examples of the errors on ali and aws that you've seen?

In general, the updates look okay but right now I cannot tell what error you've seen that creates the need for this PR.

Additionally, since this only happens with specific disks I'm wondering if this should be enough reason to actually add an integration test this..

fmoehler:
Hi, 

in the linked [issue](https://github.com/cloudfoundry/bosh-agent/issues/310) the errors are described. Shall I provide some more examples?
If needed, I could also add an integration test for this scenario.

nouseforaname:
Ah nice. I think I'd be okay if the tests you already added got updated with a few real live ali / aws paths so we leave a track of how these look RN..

The reason is that I don't know if this
a) never worked for these types of disks
b) something in ali / aws changed?

rkoster (approved):

rkoster (changes requested):
Had some internal discussion and posted a follow up question here (the issue context, so lets discuss the problem a bit more, and figure out where to fix it): https://github.com/cloudfoundry/bosh-agent/issues/310#issuecomment-1544481176

rkoster:
Closing in favor of https://github.com/cloudfoundry/bosh-agent/pull/311

Issue #310, opened 2023-05-04
Device path resolution times out for aws and ali


fmoehler:
We found that the resolution of the device path by ID times out on ali and aws. This is due to a mismatch in the disk ID and the symlink in /dev/disk/by-id.
Please see below the agent logs for aws:
```
2023-05-03_16:01:00.86472 [File System] 2023/05/03 16:01:00 DEBUG - Glob '/dev/disk/by-id/*vol-0762ac95a0e2714e6'
2023-05-03_16:01:00.86486 [virtioDevicePathResolver] 2023/05/03 16:01:00 DEBUG - Failed to get device real path by disk ID: 'vol-0762ac95a0e2714e6'. Error: 'Timed out getting real device path for 'vol-0762ac95a0e2714e6'', timeout: 'true'
2023-05-03_16:01:00.86487 [virtioDevicePathResolver] 2023/05/03 16:01:00 DEBUG - Using mapped resolver to get device real path
2023-05-03_16:01:00.86487 [File System] 2023/05/03 16:01:00 DEBUG - Checking if file exists /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol0762ac95a0e2714e6 
```
As can be seen the disk id contains the "-" character, however the symlink does not contain it (last log entry at the end)
The issue for ali is slightly different.
First the agent logs:
```
2023-05-03_16:05:38.53543 [File System] 2023/05/03 16:05:38 DEBUG - Glob '/dev/disk/by-id/*d-gw887i4b4bb3n9acrozc'
2023-05-03_16:05:38.53552 [virtioDevicePathResolver] 2023/05/03 16:05:38 DEBUG - Failed to get device real path by disk ID: 'd-gw887i4b4bb3n9acrozc'. Error: 'Timed out getting real device path for 'd-gw887i4b4bb3n9acrozc'', timeout: 'true'
```
and here the content of the /dev/disk/by-id folder:
![image](https://user-images.githubusercontent.com/101196374/236235970-3f60a5bb-c8c3-40a5-aefd-eb113ce04d11.png)

We propose the following change: https://github.com/cloudfoundry/bosh-agent/pull/309 

This PR removes everything from the DiskID before and including the "-" character, if it exists.

rkoster:
Why is the cpi passing down the `d-`? Wouldn't it be better if that would have been stripped in the cpi instead of doing this in the agent?

klakin-pivotal:
[...]

PR #311, opened 2023-06-15
Add StripVolumeRegex configuration property

ten4o:
Add a `StripVolumeRegex` configuration property that will be used to strip some part of `disk-id` (aka volume ID) before device path resolve step, as proposed by the Working Group:

_Create a configuration property for the bosh agent to configure a strip-volume-regex,
which would allow specifying a regex prefix per IaaS in the bosh stemcell builder [here](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/stemcell_builder/stages/bosh_alicloud_agent_settings/apply.sh#L11-L1) (in the case of Ali Cloud)._

The potential benefit is that this could solve issue #310.

### DIFF
--- a/infrastructure/devicepathresolver/id_device_path_resolver.go
+++ b/infrastructure/devicepathresolver/id_device_path_resolver.go
@@ -26,7 +26,7 @@ func NewIDDevicePathResolver(
 	fs boshsys.FileSystem,
 	stripVolumeRegex string,
 ) DevicePathResolver {
-	return idDevicePathResolver{
+	return &idDevicePathResolver{
 		diskWaitTimeout:     diskWaitTimeout,
 		udev:                udev,
 		fs:                  fs,
@@ -35,7 +35,7 @@ func NewIDDevicePathResolver(
 	}
 }
 
-func (idpr idDevicePathResolver) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
+func (idpr *idDevicePathResolver) GetRealDevicePath(diskSettings boshsettings.DiskSettings) (string, bool, error) {
 	if diskSettings.ID == "" {
 		return "", false, bosherr.Errorf("Disk ID is not set")
 	}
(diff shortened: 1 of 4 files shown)


beyhan (changes requested):

beyhan (approved):

beyhan:
The change looks good. One open question is whether we need integration tests for it. I couldn't find any ITs for this kind of functionality.

ramonskie:
as we currently only do bats testing on gcp. i don't know if this pr actually solves it for this.
but if so we can maby fit is somewhere here https://github.com/cloudfoundry/bosh-acceptance-tests/blob/master/spec/bat/bosh_helper_spec.rb#L29

beyhan:
We want to make use of this only on AWS and Ali for now. On GCP stripping disk IDs isn't required. I was thinking for agent integration tests and not bosh.

rkoster:
There are some integration test helpers like [AttachDevice](https://github.com/cloudfoundry/bosh-agent/blob/main/integration/test_environment.go#L321) which could be helpful when adding an integration test.

andinod:
Hi, 

one question, is this change considering the Openstack Env?, because I have a similar issue described here [https://github.com/cloudfoundry/bosh-agent/issues/310#issuecomment-1611380262](https://github.com/cloudfoundry/bosh-agent/issues/310#issuecomment-1611380262) and I would like to know if I need to wait or there is a workaround that I have to look for.

Thanks in advance,

David

beyhan:
Hi @andinod,
this change will enable us to configure a prefix which could be removed from the disk ID on the different IaaS providers. That means configuration like [this](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/stemcell_builder/stages/bosh_alicloud_agent_settings/apply.sh#L11-L13) will be required for an IaaS provider so that this can have an affect as discussed [here](https://github.com/cloudfoundry/bosh-agent/pull/309#issuecomment-1572247146).

beyhan:
Regarding the IT discussion:
I think that it make sense to add a test case to the [AssociateDisk](https://github.com/cloudfoundry/bosh-agent/blob/main/platform/linux_platform_test.go#L3575-L3605) tests in the [linux_platform_test.go](https://github.com/cloudfoundry/bosh-agent/blob/main/platform/linux_platform_test.go). The settings should look like [this](https://github.com/cloudfoundry/bosh-agent/blob/main/integration/assets/file-settings-agent.json) so that the right provider is selected [here](https://github.com/cloudfoundry/bosh-agent/blob/main/platform/provider.go#L116-L121). Additionally, we should configure a strip value and verify that the ID resolution worked. @jpalermo, @rkoster  what opinion on this? Are the agent integration tests executable locally because our [agent dev setup](https://github.com/cloudfoundry/bosh-agent/blob/main/docs/dev_setup.md) is not clear about this and I didn't have time to try it?

rkoster (approved):

beyhan (changes requested):
Please try to add the integration test as I described it in my previous comment.

ten4o:
@beyhan After examining `AssociateDisk` tests, it seems that those are the unit tests of `linux_platform.go` - all test cases in that file are using `FakeDevicePathResolver` which will not call our code in `GetRealDevicePath`. It seems that there is not much value in hacking together additional test cases in `AssociateDisk` when we can already add similar test cases in `identity_device_path_resolver_test.go`.

beyhan:
Ok, I looked into the wrong tests. Looking into the right tests I think an integration test will be challenging to implement here because the agent handles the resolution internally during the [update_setttings](https://github.com/cloudfoundry/bosh-agent/blob/main/agent/action/update_settings.go#L44-L81) action. The only way to validate from outside that the [id_device_path_resolver.go](https://github.com/cloudfoundry/bosh-agent/blob/main/infrastructure/devicepathresolver/id_device_path_resolver.go) provider was used, is via logs which is error prone.

beyhan (approved):
We agreed to go without integration tests.

Issue #313, opened 2023-09-28
Persistent disk resize fails to handle linux sparse files, resulting in filling up target persistent disk


gberche-orange:
## Problem statement

Given a bosh persistent disk contains a [Linux sparse file](https://en.wikipedia.org/wiki/Sparse_file) such as the one created by the following command
```bash
mkdir -p /var/vcap/store/disk-resize-source-dir
cd /var/vcap/store/disk-resize-source-dir

# create an empty sparse file of 5MB
truncate -s 5M /var/vcap/store/disk-resize-source-dir/a-5MB-sparse-file.txt

# check it actually occupies 0 bytes on the device
ls --size --human /var/vcap/store/disk-resize-source-dir/a-5MB-sparse-file.txt
0 /var/vcap/store/a-5MB-sparse-file.txt

# apparent size is 5 MB
ls -l --human /var/vcap/store/disk-resize-source-dir/a-5MB-sparse-file.txt
-rw-r--r-- 1 root root 5.0M Sep 28 14:09 /var/vcap/store/disk-resize-source-dir/a-5MB-sparse-file.txt
```
When the bosh agent resizes a persistent disk a copies data using
https://github.com/cloudfoundry/bosh-agent/blob/a76e83b6ffd138901a10dd44c3ffc89156d33203/platform/linux_platform.go#L1397

```bash
mkdir -p /var/vcap/store/disk-resize-target-dir 
(tar -C /var/vcap/store/disk-resize-source-dir --xattrs --xattrs-include=*.* -cf - .) | (tar -C /var/vcap/store/disk-resize-target-dir --xattrs --xattrs-include=*.* -xpf -)
```

then the real size on device is the full reserved space instead of being empty (i.e; having holes detected)

```bash
ls --size --human /var/vcap/store/disk-resize-target-dir/
total 5.0M
5.0M a-5MB-sparse-file.txt

# apparent space remains 5MB
ls -l --human /var/vcap/store/disk-resize-target-dir/
total 5.0M
-rw-r--r-- 1 root root 5.0M Sep 28 14:09 a-5MB-sparse-file.txt
``` 

See full related story at https://github.com/orange-cloudfoundry/paas-templates/issues/2060

Associated slack discussion https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1695909438966129

## Proposed fix: add the `--sparse` to the gnu tar command

https://www.gnu.org/software/tar/manual/html_node/sparse.html
> ‘--sparse’
[...]

PR #314, opened 2023-09-28
Support Linux sparse files

gberche-orange:
Pass the --sparse to the tar create archive command used for disk resize. This option will detect holes in the file and only copy non-empty data. The --sparse option is omitted on the tar extraction command where it has no effect.

Fixes https://github.com/cloudfoundry/bosh-agent/issues/313


aramprice (commented):
lgtm

jpalermo (approved):
👍

jpalermo:
Merging, assuming nothing breaks, it should go out in the next jammy release.

Thanks!

PR #322, opened 2024-03-19
fix(bosh-agent compile): fingerprint and dependency fields do not match source tarball

crhntr:
fixes: https://github.com/cloudfoundry/bosh-agent/issues/323

### DIFF
--- a/releasetarball/compile_test.go
+++ b/releasetarball/compile_test.go
@@ -54,8 +54,7 @@ var _ = Describe("NewCompiler", func() {
 			Expect(setupErr).NotTo(HaveOccurred())
 
 			d := directories.NewProvider(temporaryDirectory)
-			err := os.MkdirAll(d.BlobsDir(), 0o766)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(os.MkdirAll(d.BlobsDir(), 0o766)).To(Succeed())
 
 			result, err := releasetarball.NewCompiler(d)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


releasetarball/compile_test.go
```diff
@@ -182,14 +182,29 @@ var _ = Describe("Compile", func() {
 	})
 
 	When("compiling a release with multiple source packages", func() {
+		var multiplePackageCompiler *fakes.Compiler
 		BeforeEach(func() {
 			sourceTarballPath = filepath.Join("testdata", "log-cache-release-3.0.9.tgz")
+
+			multiplePackageCompiler = new(fakes.Compiler)
+			Expect(os.WriteFile(filepath.Join(d.BlobsDir(), "golang-1.20-linux-blob"), []byte("golang-1.20-linux-blob"), 0644)).NotTo(HaveOccurred())
```
aramprice, line 190: It may be clearer to write this as `Expect(s.WriteFile(...)).To(Succeed())`.

aramprice (approved):

releasetarball/compile_test.go
crhntr: Cool! That is clearer. I forgot that matcher was a thing. I generally avoid Gomega/Gingo on my own projects so I forget how expressive it can be.

crhntr: I pushed the refactor. I found a few other assertions that could use the `To(Succeed())` pattern and I converted the block you mentioned into a loop.

aramprice: Cool, thanks!

jpalermo (approved):
👍

PR #325, opened 2024-04-24
Adding support for EFI based stemcells.

lnguyen:
GCP repartition the root disks. It doesn't take in account if a system has more then 1 partition.

This has been tested with EFI stemcell and non EFI stemcell

### DIFF
--- a/platform/disk/root_device_partitioner.go
+++ b/platform/disk/root_device_partitioner.go
@@ -104,6 +104,7 @@ func (p rootDevicePartitioner) GetDeviceSizeInBytes(devicePath string) (uint64, 
 	}
 
 	firstPartitionInfo := strings.Split(partitionInfoLines[1], ":")
+	// If EFI partition is used, use the second partition as it is root partition
 	if firstPartitionInfo[4] == "fat16" {
 		firstPartitionInfo = strings.Split(partitionInfoLines[2], ":")
 	}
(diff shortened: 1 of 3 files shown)


platform/disk/root_device_partitioner.go
```diff
@@ -170,17 +169,26 @@ func (p rootDevicePartitioner) GetPartitions(devicePath string) (
...
 
+		partitionType := PartitionTypeUnknown
+		if partitionInfo[4] == "ext4" || partitionInfo[4] == "xfs" {
+			partitionType = PartitionTypeLinux
+		} else if partitionInfo[4] == "linux-swap(v1)" {
+			partitionType = PartitionTypeSwap
+		} else if partitionInfo[4] == "fat16" {
+			partitionType = PartitionTypeEFI
+		}
+
```
beyhan, line 180: I couldn't follow here whether this is an additional improvement or related to the EFI based stemcell change. The same with the new if in line 107 to 109. Maybe an additional comment could help here.

lnguyen: I added comments. Let me know if it is helpful. It is used to determine if we are using EFI to check 2nd partition instead of 1st. We only assumed 1st before.

beyhan: @lnguyen Thanks! It helps me at least.

aramprice (approved):

PR #327, opened 2024-04-30
Support systemd for monit

lnguyen:
With the move to Ubuntu Noble. sv is no longer used. We use systemd to manage monit. This change moves agent setting from DNSResolverType to a generic ServiceManager. The monit job supervisor now detect and uses correct one.


jpalermo (approved):
👍

ramonskie (approved):
have not reviewed all the code.
but i tested this with bats tests and works as expected.

PR #332, opened 2024-06-17
Skip NATS firewall creation when cgroupv2 enabled

ramonskie:
this will add the following nftables rules 
when cgroupv2 are enabled on a linux system
```
socket cgroupv2 level 2 "system.slice/bosh-agent.service" ip daddr 1.2.3.4 tcp dport 1234 log prefix "Matched cgroup bosh-agent nats rule: " accept
meta skuid 0 ip daddr 1.2.3.4tcp dport 1234 log prefix "Matched skuid director nats rule: " accept
ip daddr 1.2.3.4 tcp dport 1234 log prefix "dropped nats rule: " drop
```


platform/net/firewall_provider_test.go
```diff
@@ -22,3 +24,45 @@ var _ = Describe("SetupFirewall Linux", func() {
...
+						Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Did not expect error for mbus: %s", test.mbus))
+					}
+				}
+			})
+		})
+	})
+
+	Describe("Adding Nftables Rules", func() {
+		Context("With different rule expressions", func() {
+			FIt("Should add rules for cgroup match", func() {
```
aramprice, line 56: `FIt` should be changed back to `It` before merging.

platform/net/firewall_provider_test.go
```diff
@@ -22,3 +24,45 @@ var _ = Describe("SetupFirewall Linux", func() {
...
+				// this tests needs the bosh-agent cgroup to be created and nftables to be installed
+				err := SetupNFTables("1.2.3.4", "1234")
+				Expect(err).ToNot(HaveOccurred(), "Failed to setup nftables")
```
aramprice, line 59: I think these linter issuec be avoided by moving linux specific tests to a `_linux` suffixed file.

Alternately we can drop macOS linting which is probably a more reasonable/sustainable decision.

ramonskie: i agree to remove the macOS linting. as it has no added value.

ramonskie: on the otherhand it makes sense to give a linux suffix as we are actaully testing `firewall_provider_linux.go`

ramonskie:
currenly the test will not work as we don't test it on a stemcell.
as this test assumes it has a bosh-agent and systemd available
so the test should be improved. any suggestions?

platform/net/firewall_provider_linux.go
```diff
@@ -45,8 +54,13 @@ func SetupNatsFirewall(mbus string) error {
...
 		return nil
 	}
+	// NOBLE_TODO: check if warden does not hit this cgroup v2 code path
+	var cgroupV2 bool
+	if cgroups.Mode() == cgroups.Unified {
+		cgroupV2 = true
+	}
 	_, err := cgroups.V1()
-	if err != nil {
+	if err != nil && !cgroupV2 {
```
aramprice, line 63: It might make this method less confusing to move the 
```golang
	_, err := cgroups.V1()
	if err != nil {
	//snip
	}
```

section down into the `SetupIptables()` func. This would allow the check at the end of this func to be:

```golang
	if cgroups.Mode() == cgroups.Unified {
		return SetupNFTables(host, port)
	} else {
		return SetupIptables(host, port, addr_array)
	}
```

... and for `cgroups.V1()` to only be checked inside `SetupIptables()` where it is required.

aramprice:
For testing, if we are relying on the CFF Concourse instances, we may need to wait for a cgroups v2 stemcell to become available and for Concourse to be able to consume it for the cgroup part. Unsure about `systemd` and whether that's present in a Concourse container.

jpalermo:
We talked in the FIWG meeting about having the tests conditional on the cgroups currently enabled. So for now, in the bosh-agent pipelines, the cgroupsv2 code path wouldn't actually get tested. But as soon as we start shipping Noble and add a job for the pipeline, the cgroupsv2 tests would be active when running against Noble.

rkoster:
@ramonskie did you have a chance to take a look at ^

rkoster:
Once CI is green again @ramonskie is gonna look at this (as discussed during working group meeting).

platform/net/firewall_provider_linux.go
nader-ziada: change already made

jpalermo:
Hey @ramonskie and @rkoster, some of the new tests are still failing because they're trying to call `iptables` during the `SetupNatsFirewall(test.mbus)` call in the tests.

Since the unit tests are running on jammy, they're not using the nftables branch.

We'd probably need to either delete that test, or inject enough dependencies in the unit tests to prevent them from actually doing anything.

But larger question, can we just delete all this code?

The firewall rules are designed to prevent workloads from talking to NATs because the NATs creds used to be accessible from the IaaS metadata endpoints. But we've switched to using temporary NATs creds, so that's no longer an attack vector.

This does provide an additional layer of security, but it's a LOT of complexity to maintain for a "second fence".

We should discuss at the FIWG meeting tomorrow.

beyhan:
The decision was made to not implement this logic in Nobel because it doesn't bring more value from security perspective.

aramprice (approved):

jpalermo:
@beyhan, I added you as a reviewer because you said you were going to check if you all had any security concerns with this.

ystros (approved):

jpalermo:
@a-hassanin , just assigning you to check if you have any security concerns about this.

aramprice:
The following patch should resolve the linter error:

```patch
diff --git a/platform/net/firewall_provider_linux.go b/platform/net/firewall_provider_linux.go
index 306d8938..9298d665 100644
--- a/platform/net/firewall_provider_linux.go
+++ b/platform/net/firewall_provider_linux.go
@@ -6,17 +6,15 @@ package net
 import (
 	"errors"
 	"fmt"
-	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 	"net"
 	gonetURL "net/url"
 	"os"
 	"strings"
-	// NOTE: "cgroups is only intended to be used/compiled on linux based system"
-	// see: https://github.com/containerd/cgroups/issues/19
-	"github.com/containerd/cgroups"
-	"github.com/opencontainers/runtime-spec/specs-go"
 
+	bosherr "github.com/cloudfoundry/bosh-utils/errors"
+	"github.com/containerd/cgroups" // NOTE: linux only; see: https://github.com/containerd/cgroups/issues/19
 	"github.com/coreos/go-iptables/iptables"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
 const (
```

rkoster:
@beyhan any news on the security front?

beyhan:
I'm waiting for feedback here.

a-hassanin:
Mentioning the feature PR https://github.com/cloudfoundry/bosh/pull/2417 as it is relevant for the decision here.

a-hassanin:
It looks Ok from our side we can merge it. We will need to revisit this from SAP side when we want to use Noble bec. as I saw we need a PoC to make sure it will work for AliCloud. But this we will plan later.

a-hassanin (approved):

aramprice (approved):

PR #338, opened 2024-11-10
Noble VMs on AWS don't have /dev/sdX devices for some reason.

jpalermo:
We'd like to use the idDevicePathResolver and disk hints instead, but we don't actually know the volume id for the ephemeral disk until the instance is created at which point it's too late to put the disk into into the metadata because the VM has to be in a stopped state to do that.

So instead, we just guess here and hope the nvme devices match up. Which isn't much worse than what we were doing before, but AWS was also just sort of hoping that the device would end up on /dev/sdb.

### DIFF
--- a/infrastructure/devicepathresolver/mapped_device_path_resolver.go
+++ b/infrastructure/devicepathresolver/mapped_device_path_resolver.go
@@ -1,6 +1,7 @@
 package devicepathresolver
 
 import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


lnguyen (approved):

ramonskie (approved):
copied the build agent bin it in to the stemcell and did a full test.
works as expected

PR #343, opened 2024-12-18
ID Device path resolver

ramonskie:
we noticed that the id device path resolver code path always did a fallback to other mechanisms to mount the disk.
and this was due to the fact that the `DiskID` provided by the agent never matched the actual disk.

with this change you can now add regex patterns and replacements in the agent.json to match the expected disk-id

for example AWS:
provides the disk id as `vol-04b2a40747d7e9625` but the actuall disk path is `/dev/disks/by-id/nvme-Amazon_Elastic_Block_Store_vol04b2a40747d7e9625`

and if we place the following in the agent.json
```
"DiskIDTransformPattern": "^(vol-(.+))$",
"DiskIDTransformReplacement": "nvme-Amazon_Elastic_Block_Store_vol${2}"
```
it would be transformed to `vme-Amazon_Elastic_Block_Store_vol04b2a40747d7e9625`


fmoehler (approved):

jpalermo (approved):
👍

platform/linux_platform.go
```diff
@@ -84,9 +84,10 @@ type LinuxOptions struct {
 	// possible values: systemd, ""
 	ServiceManager string
 
-	// Regular expression specifying what part of volume ID to strip.
-	// possible values:  valid RE2 regex e.g. "^vol-", "" (default is not to strip)
-	StripVolumeRegex string
+	// Regular expression specifying what part of disk ID to strip and transform
+	// example: "pattern": "^(disk-.+)$", "replacement": "google-${1}",
+	DiskIDTransformPattern     string
+	DiskIDTransformReplacement string
```
jpalermo, line 90: FYI, bosh.io [links to](https://bosh.io/docs/agent-cpi-interactions/#agent-json) golang docs that document these flags

anshrupani (approved):

PR #344, opened 2024-12-19
Disable error messages related to networks/network interface cards

fmoehler:
Some iaas (e.g. AWS) support dual-stack network interface cards. However as bosh only supports to provide one range (either ipv4 or ipv6) in the cloud config two networks need to be defined to specify both ranges. Therefore the error message can be deleted.

### DIFF
--- a/agent/bootstrap_test.go
+++ b/agent/bootstrap_test.go
@@ -1209,14 +1209,6 @@ var _ = Describe("bootstrap", func() {
 							}`
 				})
 
-				Context("and no physical network interfaces exist", func() {
-					It("raises an error", func() {
-						err := boot.Run()
-						Expect(err).To(HaveOccurred())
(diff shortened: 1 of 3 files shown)


jpalermo (approved):

ystros (approved):
Overall looks good. 

I'm slightly hesitant about the removal of error handling when there are 0 physical network interfaces passed to the method. Going up the call stack, it looks like the function to retrieve them would have to return an empty array without an error. But it seems somewhat unrealistic that there will be 0 network devices, so it may be fine to remove that check.
https://github.com/cloudfoundry/bosh-agent/blob/29b4d4d505d55161c481844c9a87aade696290ee/platform/net/ubuntu_net_manager.go#L304-L307

Also dug up the original commit that introduced the check being removed: https://github.com/cloudfoundry/bosh-agent/commit/348a0149c1ee34988f25676ce17f07911b970320

PR #351, opened 2025-04-30
skip network resolution if prefix defined

fmoehler:


platform/linux_platform.go
```diff
@@ -1530,8 +1530,8 @@ func (p linux) DeleteARPEntryWithIP(ip string) error {
 	return nil
 }
 
-func (p linux) GetDefaultNetwork() (boshsettings.Network, error) {
-	return p.defaultNetworkResolver.GetDefaultNetwork()
+func (p linux) GetDefaultNetwork(ipv6 bool) (boshsettings.Network, error) {
+	return p.defaultNetworkResolver.GetDefaultNetwork(ipv6)
```
aramprice, line 1534: Rather than having a `bool` hear could `boshsettings.Network` be changed to include IP version information?

It seems odd for the IPv6-ness of the network to be communicated in a way that is disjoint from `Network` itself.

platform/net/ip/ip_resolver.go
```diff
@@ -20,6 +20,7 @@ func NetworkInterfaceToAddrsFunc(interfaceName string) ([]gonet.Addr, error) {
 type Resolver interface {
 	// GetPrimaryIPv4 always returns error unless IPNet is found for given interface
 	GetPrimaryIPv4(interfaceName string) (*gonet.IPNet, error)
+	GetPrimaryIPv6(interfaceName string) (*gonet.IPNet, error)
```
aramprice, line 23: See https://github.com/cloudfoundry/bosh-agent/pull/352#issuecomment-2863633395

platform/linux_platform.go
fmoehler: Hi Aram,

thanks for checking and thanks for the feedback! 
Sorry for getting back to you only after all this time.
We did check again why we implemented it this way and found that probably this config parameter is the "correct" approach. The GetDefaultNetwork returns a network. However this is not the network that is passed via user-data from the director, but the network that the bosh-agent itself "resolved" (by using `ip r` or `ip -6 r` commands). Therefore the agent somehow needs to know if it should resolve an ipv6 or ipv4 network. (it is passed all the way through till it checks for the routes here: https://github.com/cloudfoundry/bosh-agent/pull/351/files#diff-7cd9514942becfaccdb4de35dc7ca1a482011459843050cf0a1b9cd67b733baaR60-R69 ). I think this PR even fixes a bug where the network interface might have been correct, but the spec.json was always wrong, because it was never checking for the ipv6 routes.

aramprice: Hi Felix,

That explanation makes sense as to why an IP type would need to be sent down.

Would it make sense to expand `boshsettings.Network` to include both IPv4, and IPv6 information and for the agent to execute both `ip r` and `ip -6 r` so that the return value includes both bits of information? 
^ not this might be a silly suggestion, I haven't looked closely at the code.

If checking / returning both types of address doesn't make sense then I think it would be nice to change from `ipv6 bool` to some kind of constant / enum for the IP version? Something like:

```golang
// Declaring an enum type for departments
type IPAddressVersion int

const (
	IPv4 = iota
	IPv6
)

// ....

func (p linux) GetDefaultNetwork(ipVersion IPAddressVersion) (boshsettings.Network, error) {
     // ...
}
```

platform/linux_platform.go
```diff
@@ -1530,8 +1530,8 @@ func (p linux) DeleteARPEntryWithIP(ip string) error {
 	return nil
 }
 
-func (p linux) GetDefaultNetwork() (boshsettings.Network, error) {
-	return p.defaultNetworkResolver.GetDefaultNetwork()
+func (p linux) GetDefaultNetwork(is_ipv6 bool) (boshsettings.Network, error) {
```
aramprice, line 1533: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... `is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/net/arp/arping_test.go
```diff
@@ -19,7 +19,7 @@ type failingInterfaceAddress struct{}
 
 func (ia failingInterfaceAddress) GetInterfaceName() string { return "eth0" }
 
-func (ia failingInterfaceAddress) GetIP() (string, error) {
+func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error) {
```
aramprice, line 22: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... `is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/net/fakes/fake_routes_searcher.go
```diff
@@ -9,6 +9,6 @@ type FakeRoutesSearcher struct {
 	SearchRoutesErr    error
 }
 
-func (s *FakeRoutesSearcher) SearchRoutes() ([]boshnet.Route, error) {
+func (s *FakeRoutesSearcher) SearchRoutes(ipv6 bool) ([]boshnet.Route, error) {
```
aramprice, line 12: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/linux_platform_test.go
```diff
@@ -4136,15 +4136,17 @@ unit: sectors
...
-		It("delegates to the defaultNetworkResolver", func() {
-			defaultNetwork := boshsettings.Network{IP: "1.2.3.4"}
-			fakeDefaultNetworkResolver.GetDefaultNetworkNetwork = defaultNetwork
-
-			network, err := platform.GetDefaultNetwork()
-			Expect(err).ToNot(HaveOccurred())
-
-			Expect(network).To(Equal(defaultNetwork))
-		})
+		for _, value := range []bool{true, false} {
```
aramprice, line 4139: Can `value` be renamed to something that expresses intent (like whatever is chosen as the name of the `bool` where `GetDefaultNetwork()` is defined?

platform/net/ip/fakes/fake_ip_resolver.go
```diff
@@ -1,16 +1,20 @@
...
+	GetPrimaryIPInterfaceName string
+	GetPrimaryIPNet           *gonet.IPNet
+	GetPrimaryIPErr           error
+	GetPrimaryIPCalledWith    []string
 }
 
-func (r *FakeResolver) GetPrimaryIPv4(interfaceName string) (*gonet.IPNet, error) {
-	r.GetPrimaryIPv4InterfaceName = interfaceName
-	return r.GetPrimaryIPv4IPNet, r.GetPrimaryIPv4Err
+func (r *FakeResolver) GetPrimaryIP(interfaceName string, is_ipv6 bool) (*gonet.IPNet, error) {
```
aramprice, line 15: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/net/ip/interface_address_test.go
```diff
@@ -90,24 +91,44 @@ var _ = Describe("resolvingInterfaceAddress", func() {
...
+		Context("when GetIP was called with true or false", func() {
+
+			BeforeEach(func() {
+				ipResolver.GetPrimaryIPNet = &gonet.IPNet{
+					IP:   gonet.ParseIP("127.0.0.1"),
+					Mask: gonet.CIDRMask(16, 32),
+				}
+			})
+
+			for _, value := range []bool{true, false} {
```
aramprice, line 123: Can `value` be renamed to something that expresses intent (like whatever is chosen as the name of the `bool` where `GetIP()` is defined?

platform/net/ip/ip_resolver.go
```diff
@@ -18,8 +18,8 @@ func NetworkInterfaceToAddrsFunc(interfaceName string) ([]gonet.Addr, error) {
 }
 
 type Resolver interface {
-	// GetPrimaryIPv4 always returns error unless IPNet is found for given interface
-	GetPrimaryIPv4(interfaceName string) (*gonet.IPNet, error)
+	// GetPrimaryIP always returns error unless IPNet is found for given interface
+	GetPrimaryIP(interfaceName string, is_ipv6 bool) (*gonet.IPNet, error)
```
aramprice, line 22: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/net/routes_searcher_interface.go
```diff
@@ -8,11 +8,15 @@ type Route struct {
 }
 
 type RoutesSearcher interface {
-	SearchRoutes() ([]Route, error)
+	SearchRoutes(ipv6 bool) ([]Route, error)
```
aramprice, line 11: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/net/routes_searcher_interface.go
```diff
@@ -8,11 +8,15 @@ type Route struct {
...
 }
 
 const DefaultAddress = `0.0.0.0`
+const DefaultAddressIpv6 = `::`
 
-func (r Route) IsDefault() bool {
+func (r Route) IsDefault(isIpv6 bool) bool {
```
aramprice, line 17: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/net/routes_searcher_windows.go
```diff
@@ -15,15 +15,26 @@ func NewRoutesSearcher(_ boshlog.Logger, cmdRunner boshsys.CmdRunner, interfaceM
 	return windowsRoutesSearcher{interfaceManager, cmdRunner}
 }
 
-func (s windowsRoutesSearcher) SearchRoutes() ([]Route, error) {
+func (s windowsRoutesSearcher) SearchRoutes(ipv6 bool) ([]Route, error) {
```
aramprice, line 18: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

platform/windows_platform_test.go
```diff
@@ -269,15 +269,17 @@ var _ = Describe("WindowsPlatform", func() {
...
-		It("delegates to the defaultNetworkResolver", func() {
-			defaultNetwork := boshsettings.Network{IP: "1.2.3.4"}
-			fakeDefaultNetworkResolver.GetDefaultNetworkNetwork = defaultNetwork
-
-			network, err := platform.GetDefaultNetwork()
-			Expect(err).ToNot(HaveOccurred())
-
-			Expect(network).To(Equal(defaultNetwork))
-		})
+		for _, value := range []bool{true, false} {
```
aramprice, line 272: Same as above `value` should have a name that expresses intent, ideally matching the name in the signature of `GetDefaultNetwork()`

settings/settingsfakes/fake_platform_settings_getter.go
```diff
@@ -128,7 +128,7 @@ func (fake *FakePlatformSettingsGetter) GetAgentSettingsPathReturnsOnCall(i int,
 	}{result1}
 }
 
-func (fake *FakePlatformSettingsGetter) GetDefaultNetwork() (settings.Network, error) {
+func (fake *FakePlatformSettingsGetter) GetDefaultNetwork(isIpV6 bool) (settings.Network, error) {
```
aramprice, line 131: can we choose `ipv6`, `isIpv6` (or perhaps `isIPv6`)... 

`is_ipv6` seems wrong in that Golang suggest camelCase for variables, as does `IsVersion6` (see `func (ia failingInterfaceAddress) GetIP(IsVersion6 bool) (string, error)`)

aramprice (changes requested):
Overall the shape of this looks reasonable.

I'm not super comfortable with the use of a `bool` to handle the conditionality between IPv6 or IPv4. It make the code harder to read, and I think more fragile.

I think an better approach would be to have a `type` that indicates which IP version is being requested, rather than passing a boolean. 

Maybe something like the following...
```golang
type IPVersion bool // name isn't great

const (
	IPv4 IPVersion = false
	IPv6 IPVersion = true
)
```

... or perhaps

```golang
type IPVersion int64 // name isn't great

const (
	IPv4 IPVersion = iota
	IPv6
)
```

At a minimum if an bare `bool` is used I think the parameter name in all of the `func` defs should be consistent. Ideally I think some form of typing should be used so that the code is easier to read / navigate.

aramprice:
@fmoehler - thanks for the changes, this looks good.

FYI I pushed a fix for the windows lint issue: https://github.com/cloudfoundry/bosh-agent/commit/3aba0d852b7da62d8b1801d6ce714b5cc8015a4c

aramprice (approved):
This looks good, thank you for switching from `bool` to something typed.

fmoehler:
Hi @aramprice . Thanks for your review. Sorry for only getting to it now. After rebasing the branch unfortunately there were issues with the iptables.Protocol type on windows. I have therefore switched to a custom type in the "ip" package. I hope that is still ok.

PR #352, opened 2025-05-05
check for prefix in service.go

dudejas:


aramprice:
Instead of adding `GetPrimaryIPv6` and `SearchIPv6Routes`, would it be possible to have generic IP version agnostic public methods (ex: `SearchRoutes` and `GetPrimaryIP`) which call into IP version specific non-exported functions?
> ex:
> `GetPrimaryIP` internally calls either `getPrimaryIPv4` or `getPrimaryIPv6`
> `SearchRoutes` internally calls either `searchRoutesIPv4` or `searchRoutesIPv6`

Perhaps we _do_ need to expose both `IPv6` and `IPv4` variants for get-primary-ip and search-routes, if that is the case then it would be good if the variants were named consistently:
> ex:
> `GetPrimaryIPv4` , `GetPrimaryIPv6`
> `SearchRoutesIPv4`, `SearchRoutesIPv6`

PR #354, opened 2025-06-05
Fix temp-path related failures.

ragaskar:
- The fix in bd3673037 (which addresses Windows Update changes that cause the root disk to run out of space in some deployment scenarios) failed for some deployments (with a "cannot find path" error originating from groot). We suspect (but are having trouble validating) that the problem is incomplete percolation of the SystemTemp env var.
- This commit reverts bd3673037 in favor of sym-linking C:\Windows\SystemTemp, another way of achieving the desired outcome: making sure lifecycle scripts write temp data to the ephermeral disk. A deployment configuration that failed with the changes in bd3673037 completes successfully with this change.
- A possible risk here is anything that might have previously _explicitly_ written to C:\Windows\SystemTemp will now appear instead in C:\var\vcap\data\tmp, which may have different permissions -- an incremental change to bd3673037, which would have seen processes previously acquiring a temp dir from system calls writing to C:\var\vcap\data\tmp. Given that we would expect processes prior to the Windows Update change to _also_ have been writing to C:\var\vcap\data\tmp (via the TEMP and TMP env vars), this would not seem to overall change current behavior, although could be surprising with regards to future expectations around temp dirs.

*please note: NOT YET READY TO BE MERGED. I am PR-ing this in order to take advantage of Github actions testing (i am seeing possibly spurious fails when running tests on my local machine). *


aramprice (approved):
this seems like a reasonable shift in how we're setting / preparing temp on windows.

ragaskar:
I see this is failing a bosh-agent integration test, will fix.

PR #355, opened 2025-07-17
Enable IPv6 in the kernel if there are any Ipv6 dhcp interfaces

fmoehler:
When IPv6 addresses are resolved via DHCP, e.g. on AWS, IPv6 is not enabled in the Kernel, because only static interface configurations are considered in the check so far. This PR expands the check to also consider dynamic interface configurations

### DIFF
--- a/platform/net/ubuntu_net_manager_ipv6_test.go
+++ b/platform/net/ubuntu_net_manager_ipv6_test.go
@@ -78,18 +78,18 @@ var _ = Describe("UbuntuNetManager (IPv6)", func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			interfaceAddrsProvider.GetInterfaceAddresses = []boship.InterfaceAddress{
-				boship.NewSimpleInterfaceAddress("ethstatic1", "2601:646:100:e8e8::103"),
+				boship.NewSimpleInterfaceAddress("ethstatic1", "2001:db8::103"),
 				boship.NewSimpleInterfaceAddress("ethstatic2", "1.2.3.4"),
-				boship.NewSimpleInterfaceAddress("ethstatic3", "2601:646:100:eeee::10"),
+				boship.NewSimpleInterfaceAddress("ethstatic3", "3fff::100"),
 			}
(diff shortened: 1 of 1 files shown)


platform/net/ubuntu_net_manager_ipv6_test.go
```diff
@@ -105,6 +105,28 @@ var _ = Describe("UbuntuNetManager (IPv6)", func() {
 			Expect(kernelIPv6.Enabled).To(BeTrue())
 		})
 
+		It("enables IPv6 if there are any dhcp IPv6 addresses", func() {
+			dhcp1Net := boshsettings.Network{
+				Type:    "manual",
+				IP:      "2601:646:100:e8e8::103",
```
peanball, line 111: please use the documentation IPv6 range for tests. Addresses in one of:
* `2001:0DB8::/32`
* `3fff::/20`

See:
* https://datatracker.ietf.org/doc/html/rfc9637#section-abstract-1
* https://datatracker.ietf.org/doc/html/rfc9637#section-6-2.2

peanball: mostly it's best practice to use documentation CIDRs. I don't have super strong feelings about this.

a-hassanin (dismissed):
This actually serves the desired behavior that is described in the deployment config [instance_groups](https://bosh.io/docs/manifest-v2/#instance-groups) under env.bosh.ipv6 

LGTM

beyhan:
Looks good but we should extend the test with the feedback provided in https://github.com/cloudfoundry/bosh-agent/pull/355#discussion_r2215150632.

platform/net/ubuntu_net_manager_ipv6_test.go
fmoehler: Fixed with: https://github.com/cloudfoundry/bosh-agent/pull/355/commits/e508bb4ae12abdf4c296be70d5ba3571b4b74e6e

aramprice (approved):

PR #358, opened 2025-07-29
[RFC0038] Ensure all configs for a network interface are resolved

fmoehler:
We are enabling dual stack on the same nic. For this we introduced a new property [aa7ffff](https://github.com/cloudfoundry/bosh/pull/2611/commits/aa7ffffd4625060a76f83b5664b39cfa225816d1) in the director which will allocate ips of different bosh networks to the same nic.

If multiple networks are allocated to the same nic, the logic in the agent needs to be adapted to also consider multiple networks for its dhcp configuration. If this is not done, this can lead to issues where e.g. ipv6 is mistakenly not activated in the kernel.

Pair: @ivaylogi98

### DIFF
--- a/settings/settings.go
+++ b/settings/settings.go
@@ -378,10 +378,10 @@ func (n Network) IsDefaultFor(category string) bool {
 }
 
 func (n Networks) NetworksForMac(mac string) []Network {
-	var networks []Network // Initialize an empty slice of Network
+	var networks []Network
 	for _, network := range n {
 		if network.Mac == mac {
-			networks = append(networks, network) // Append matching networks
+			networks = append(networks, network)
 		}
 	}
 	if len(networks) == 0 {


fmoehler:
This is ready for review

aramprice (dismissed):

settings/settings.go
```diff
@@ -377,14 +377,18 @@ func (n Network) IsDefaultFor(category string) bool {
 	return stringArrayContains(n.Default, category)
 }
 
-func (n Networks) NetworkForMac(mac string) (Network, bool) {
-	for i := range n {
-		if n[i].Mac == mac {
-			return n[i], true
+func (n Networks) NetworksForMac(mac string) []Network {
+	var networks []Network // Initialize an empty slice of Network
```
Alphasite, line 381: nit: these comments can probably be removed; they're easily inferred from the code.

fmoehler: Totally right! Deleted with [cf6df3f](https://github.com/cloudfoundry/bosh-agent/pull/358/commits/cf6df3f31e172a330a6bea2ab87ca717581f7cef)

PR #364, opened 2025-09-02
"Fix" bin/build for Apple Silicon.

ragaskar:
Any reason not to make this change?


rkoster:
Why is it required to enable CGO on apple silicon? Could you provide a bit more context?

rkoster (approved):

PR #371, opened 2025-10-06
unit test fixes

julian-hj:
- compare case insenstively when testing paths in case the OS has C:/WINDOWS instead of C:/Windows
- add a 2019 worker tag to make sure we are running on the right concourse worker

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -237,7 +237,7 @@ jobs:
           input_mapping:
             bosh-agent: bumped-bosh-agent
           tags:
-            - windows-nimbus
+            - windows-2019
       - put: bosh-agent
         params:
           repository: bumped-bosh-agent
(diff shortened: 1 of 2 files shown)


KauzClay (dismissed):

ci/pipeline.yml
```diff
@@ -238,6 +238,7 @@ jobs:
             bosh-agent: bumped-bosh-agent
           tags:
             - windows-nimbus
+            - windows-2019
```
selzoc, line 241: I think you want to remove the `windows-nimbus` tag if you only want it to run on the `2019` worker

selzoc (changes requested):

ci/pipeline.yml
julian-hj: We believe that tags are ANDed, so we’d only ever pick workers with both tags. Does that not match your understanding?

julian-hj: Or you just think the windows-nimbus tag is superfluous?

selzoc: > Or you just think the windows-nimbus tag is superfluous?

yup

julian-hj: Done.

KauzClay (approved):

selzoc (approved):

PR #430, opened 2026-05-15
improve error handling around disk partitioning

mkocher:
The test cleanup code was trying to unmount c:\, this should catch the actual failure instead of falling through to the cleanup code.


selzoc:
Was this not superceded by https://github.com/cloudfoundry/bosh-agent/pull/432 ?

PR #438, opened 2026-06-01
Set command to spawn with lower priority in GenericScript Run method

### LINKED ISSUE #337: Lifecycle hooks can make the agent unresponsive
Bosh-agent itself is already running with higher priority than BOSH/monit jobs to mitigate CPU-intensive workloads blocking the agent <-> director communication, see https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/00054bd98693465dd75eda1f12a7326fc5191804 .

However, as it seems lifecycle hooks like pre-start scripts can as well have the same negative effect on the communication with the director because they are started by the bosh-agent itself and hence run with the same priority. At least this is my assumption because I wasn't able to find a line of code that lowers that priority and looking at a VM while it is running a pre-start reveals that the pre-start script with all sub-processes runs with the same priority as the agent.

In our case cloning a lot of data from the remaining part of a BOSH-managed PostgreSQL cluster can trigger this issue inconsistently, which in extreme situations extends downtime unnecessarily because the bosh task itself errors with an agent timeout and the pre-start has to run from scratch again.

Of course as a quick mitigation we could for example renice the priority in our pre-start script. Yet I would see benefit as well as consistency and hence predictability if bosh agent starts external scripts/binaries with lower priority than itself.

neddp:
Lifecycle scripts (drain, pre-start, post-start, post-deploy, etc.) are executed by the agent and inherit its scheduling priority. When a script is CPU-intensive (e.g. cloning large amounts of data from a database cluster), it can starve the agent's own event loop, causing the director to time out with an agent-unreachable error - even though the agent is technically alive and the script is making progress. The deployment then fails and the script has to run from scratch.

The stemcell already runs the agent at nice -15 (see [bosh-linux-stemcell-builder@00054bd](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/00054bd98693465dd75eda1f12a7326fc5191804)) to give it priority over BOSH-managed jobs (nice 0). However, lifecycle scripts spawned by the agent inherit that -15 priority, defeating the purpose.

This PR sets `SpawnWithLowerPriority = true` on the `Command` struct for all lifecycle scripts, so they run at a lower priority than the agent itself:

- **Linux:** child process nice = parent nice + 5 (capped at 19). With the agent at -15, scripts get nice -10 - still above normal jobs (0), but below the agent.
- **Windows:** child process priority class is set to `BelowNormal`.

The priority logic itself lives entirely in bosh-utils ([#142](https://github.com/cloudfoundry/bosh-utils/pull/142)), which adds the `SpawnWithLowerPriority` field to `boshsys.Command` with platform-specific priority logic inlined directly (no external dependency, as suggested by [@rkoster](https://github.com/rkoster)).

## References

- Fixes #337
- Depends on cloudfoundry/bosh-utils#142

### DIFF
--- a/agent/script/generic_script.go
+++ b/agent/script/generic_script.go
@@ -84,6 +84,7 @@ func (s GenericScript) Run() error {
 	command := cmd.BuildCommand(s.path)
 	command.Stdout = stdoutFile
 	command.Stderr = stderrFile
+	command.SpawnWithLowerPriority = true
 
 	for key, val := range s.env {
 		command.Env[key] = val


neddp:
Waiting for a new bosh-utils tag.

beyhan:
merging as bosh-utils update in bosh-agent happend.

rkoster (commented):
https://github.com/cloudfoundry/bosh-utils/commit/2093ad1013f7fae6e5bf6528483a119f92d84077 
https://github.com/cloudfoundry/bosh-agent/blob/main/go.mod#L12

PR #446, opened 2026-07-07
Prevent duplicate default routes on multi-NIC DHCP VMs

Alphasite:
## Summary

- Adds `IsDefaultForGateway bool` to `DHCPInterfaceConfiguration`, populated from `networkSettings.IsDefaultFor("gateway")`.
- Emits `UseRoutes=no` in the `[DHCP]` section of generated `systemd-networkd` unit files for DHCP interfaces whose BOSH network does **not** have `default: [gateway]`.
- Prevents secondary DHCP NICs on multi-NIC VMs from installing a competing default route alongside the primary NIC.

## Problem

On multi-NIC VMs (e.g. Diego Cells on AWS with a dedicated NFS NIC), every DHCP interface gets an identical `systemd-networkd` config that allows the DHCP client to install a default route. The secondary NIC's DHCP-assigned default route conflicts with the primary NIC's route, causing intermittent routing failures.

## Prior art — already handled for static networks, missed for DHCP

The "is this NIC the default gateway?" logic already existed for **static** networks:

- `StaticInterfaceConfiguration` has carried `IsDefaultForGateway bool` for some time.
- `writeStaticInterfaceConfiguration` already gates two keys on that flag:
  - `Gateway=` in `[Network]` is **only written** when `IsDefaultForGateway == true` — so secondary static NICs never get a static default route written into their unit file.
  - `Broadcast=` in `[Address]` is similarly suppressed for non-gateway static NICs.

For **DHCP** networks the equivalent was never implemented:

- `DHCPInterfaceConfiguration` had no `IsDefaultForGateway` field.
- `writeDynamicInterfaceConfiguration` always emitted the `[DHCP]` section without `UseRoutes=no`, so the DHCP client on every NIC was free to install a default route from the DHCP server's router option — regardless of whether that NIC was designated as the gateway default in the BOSH manifest.

This PR closes the gap by mirroring the static-network treatment in the dynamic-network path.

## Changes

| File | Change |
|---|---|
| `platform/net/interface_configuration_creator.go` | Add `IsDefaultForGateway bool` field to `DHCPInterfaceConfiguration` + populate it |
| `platform/net/ubuntu_net_manager.go` | Conditionally write `UseRoutes=no` in `[DHCP]` section |
| `platform/net/interface_configuration_creator_test.go` | New tests: `IsDefaultForGateway=true` and `IsDefaultForGateway=false` cases |
| `platform/net/ubuntu_net_manager_test.go` | Update existing tests + add multi-NIC and single-NIC-no-explicit-gateway test cases |

## Directive choice: `UseRoutes=no` vs `UseGateway=no`

During development, `UseGateway=no` was evaluated as a narrower alternative — it suppresses only the default gateway route while still allowing other DHCP-learned classless static routes to be installed. However, live testing on the target stemcell revealed a compatibility issue:

> **`UseGateway=` was introduced in systemd 250. Ubuntu 22.04 Jammy ships systemd 249.**

When `UseGateway=no` was written to the unit file and `networkctl reload` was called, the systemd-networkd journal logged:

```
systemd-networkd[886]: /etc/systemd/network/10_eth1.network:12:
  Unknown key name 'UseGateway' in section 'DHCP', ignoring.
```

The secondary NIC's default route was not removed. `UseRoutes=no` has been supported since systemd 217, works on all supported stemcells, and correctly prevents the duplicate default route. For secondary DHCP NICs that share a subnet with the primary NIC, all DHCP-learned routes would be duplicates anyway, so the broader suppression is the right behaviour.

The code carries a comment explaining this trade-off and noting when `UseGateway=no` could replace it (i.e. once the minimum supported stemcell ships systemd 250+).

## Live test evidence

Tested on OpsManager 3.3.4 / AWS us-west-1, stemcell `bosh-aws-xen-hvm-ubuntu-jammy-go_agent/1.1250` (systemd 249):

- Two-NIC VM: `first-network` (eth0, `default: [gateway, dns]`) + `second-network` (eth1, no default)

**Before patch** — both files identical, two default routes:
```
default via 10.0.16.1 dev eth1 proto dhcp src 10.0.16.14 metric 1024
default via 10.0.16.1 dev eth0 proto dhcp src 10.0.16.11 metric 1024
```

**Test run 1 (Jul 8) — `networkctl reload` path:**
- Binary deployed without deleting existing files; agent overwrote only `10_eth1.network` and called `networkctl reload` (non-destructive)
- `UseRoutes=no` accepted without complaint; eth1 DHCP lease renewed in-place

**Test run 2 (Jul 9) — final branch HEAD, `systemctl restart` via stemcell script:**
- Existing network files deleted first to force full regeneration
- Agent wrote both files fresh, called `/var/vcap/bosh/bin/restart_networking` → `systemctl restart systemd-networkd`
- Both NICs re-acquired fresh DHCP leases (new networkd PID 43326)

**After patch (both runs)** — `10_eth1.network` has `UseRoutes=no`, single default route:
```
default via 10.0.16.1 dev eth0 proto dhcp src 10.0.16.11 metric 1024
```

eth1 retained its kernel-connected subnet route (`10.0.16.0/28 dev eth1 proto kernel scope link`) — traffic to/from the secondary IP remained reachable throughout.

> **Note on `systemctl restart` vs `networkctl reload`:** The stemcell's `restart_networking` script intentionally uses `systemctl restart systemd-networkd` (a full daemon restart that drops and re-acquires DHCP leases). `networkctl reload` would be non-destructive but would leave stale DHCP-installed routes in the kernel until the next lease renewal — defeating the purpose of writing `UseRoutes=no`. The full restart is the correct behaviour here.

## Test plan

- [x] `go test ./platform/net/...` passes locally (141 tests) and in CI (all platforms green)
- [x] Deploy a multi-NIC VM (primary DHCP with `default: [gateway, dns]`, secondary DHCP with no default) and verify only the primary NIC installs a default route — **confirmed live on AWS (both test runs)**
- [x] Single-NIC DHCP VM without explicit `default: [gateway]` does not get `UseRoutes=no` — **covered by new unit test**

[TNZ-116393]
[TNZ-117845]


platform/net/ubuntu_net_manager.go
```diff
@@ -531,6 +531,9 @@ func (net UbuntuNetManager) writeDynamicInterfaceConfiguration(configs DHCPInter
 	dhcpSection := &ini.Section{Name: "DHCP"}
 	dhcpSection.AddKey("UseDomains", "yes")
 	dhcpSection.AddKey("UseMTU", "yes")
+	if !configs[0].IsDefaultForGateway {
+		dhcpSection.AddKey("UseRoutes", "no")
+	}
 	file.AppendSection(dhcpSection)
```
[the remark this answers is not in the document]
colins: This suggestion from code rabbit makes sense.

platform/net/interface_configuration_creator_test.go
```diff
@@ -281,7 +281,27 @@ var _ = Describe("InterfaceConfigurationCreator", func() {
...
+					_, dhcpInterfaceConfigurations, err := interfaceConfigurationCreator.CreateInterfaceConfigurations(networks, interfacesByMAC)
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(dhcpInterfaceConfigurations).To(ConsistOf(DHCPInterfaceConfiguration{
+						Name:                "dhcp-interface-name",
+						IsDefaultForGateway: true,
+					}))
+				})
+			})
+
```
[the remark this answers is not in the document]
colins: The indentation is off; it also changes which BeforeEach setup blocks are in scope for the "no MAC address" tests.

platform/net/ubuntu_net_manager_test.go
```diff
@@ -623,6 +623,7 @@ DNS=8.8.8.8
 [DHCP]
 UseDomains=yes
 UseMTU=yes
+UseRoutes=no
```
colins: UseRoutes=no — meaning DHCP won't install a default route?
If any real deployment has a single-NIC DHCP VM that relies on DHCP for its default route but doesn't explicitly set default: [gateway] in the BOSH manifest,  this change removes the default route.

platform/net/interface_configuration_creator_test.go
```diff
@@ -281,6 +281,26 @@ var _ = Describe("InterfaceConfigurationCreator", func() {
...
+				Context("and the dhcp network has default gateway", func() {
+					BeforeEach(func() {
+						dhcpNetwork.Default = []string{"gateway"}
+						networks["foo"] = staticNetwork
+						networks["bar"] = dhcpNetwork
+						interfacesByMAC[staticNetwork.Mac] = "static-interface-name"
+						interfacesByMAC[dhcpNetwork.Mac] = "dhcp-interface-name"
+					})
+
+					It("sets IsDefaultForGateway=true on the DHCP configuration", func() {
```
colins: Add a `IsDefaultForGateway=false` test?

platform/net/ubuntu_net_manager_test.go
```diff
@@ -1011,6 +1013,63 @@ Gateway=3.4.5.6
...
+DHCP=yes
+DNS=8.8.8.8
+
+[DHCP]
+UseDomains=yes
+UseMTU=yes
+UseRoutes=no
+
 `))
 		})
```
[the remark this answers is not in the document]
colins: This makes sense

platform/net/ubuntu_net_manager_test.go
Alphasite: yes, i think thats a valid issue, i was working through this whole code path because im not a fan of `UseRoutes=no`, I think code rabbits suggestion that its excessively broad is the right one and ive changed it locally. 

But we also need to handle the case of people not specifying a default at all... i wonder if we can use the first interface as the default if no explicit default gateway is provided? Or is it better to only disable implicit default routes if any interface specifies a default gateway...

Alphasite: also very nice catch, i appreciate the rigor

colins: Can we make this seamless for the customer?  Maybe a failing test for this case then see if we can work around it?

platform/net/ubuntu_net_manager.go
```diff
@@ -531,6 +531,9 @@ func (net UbuntuNetManager) writeDynamicInterfaceConfiguration(configs DHCPInter
 	dhcpSection := &ini.Section{Name: "DHCP"}
 	dhcpSection.AddKey("UseDomains", "yes")
 	dhcpSection.AddKey("UseMTU", "yes")
+	if !configs[0].IsDefaultForGateway {
+		dhcpSection.AddKey("UseRoutes", "no")
```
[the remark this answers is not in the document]
Alphasite: Good feedback but we need to support jammy where we dont have an alternative (systemd v249 doesnt support `UseGateway`, it was added in v250)

platform/net/ubuntu_net_manager_test.go
Alphasite:
```go
isDefaultGateway := !anyIsDefaultForGateway || dynamicAddressConfigurations.IsDefaultForGateway()
```

yes, i made the change here https://github.com/cloudfoundry/bosh-agent/pull/446/changes#diff-010e8a578b784cca80730f6e56600727eede1f1fafdd186f718872bff04841e6R415 where it will only supress the implicit default gateway if there is an explicit default gateway specified.

platform/net/ubuntu_net_manager.go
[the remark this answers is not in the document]
Alphasite: Fixed — the function now receives an `isDefaultGateway bool` parameter computed at the call site as `!anyIsDefaultForGateway || dynamicAddressConfigurations.IsDefaultForGateway()`, which iterates all configs for the interface rather than assuming `configs[0]`.

platform/net/interface_configuration_creator_test.go
Alphasite: Done — added a sibling `Context("and the dhcp network does not have default gateway")` immediately after the existing true-case context in `interface_configuration_creator_test.go`.

platform/net/ubuntu_net_manager_test.go
Alphasite: The `anyIsDefaultForGateway` guard handles this — `UseRoutes=no` is only emitted when at least one NIC in the deployment explicitly declares `default: [gateway]`. A test covering the single-NIC DHCP case with no explicit default gateway is now included (`does not write UseRoutes=no for a single DHCP NIC with no explicit default gateway` in `ubuntu_net_manager_test.go`).

colins (approved):
Looks good. Thanks

PR #457, opened 2026-08-03
Speed up agent integration tests

mkocher:

The agent tests were slow and brittle. @KauzClay and I spent a bunch of time adding Noble and Resolute builds of them, and in the process saw a bunch of stuff we wanted to improve.

These changes clean up the setup and teardown, and optimize the runtime:
* Jammy: 1 Hour -> 20 Minutes
* Noble & Resolute: 47 Minutes -> 15 Minutes
(all times include the fixed ~4 minute cost of deploying a new VM)

This PR also includes documentation and changes to make it possible to run the agent suite from a workstation targeting a bosh deployed VM.

This PR is multiple commits. Reviewing by commit-by-commit may be easier. Please do not squash them if you're merging this PR.

These changes have gone green 4x in a row in a reproduction of the CI pipeline running builds for Jammy, Noble and Resolute. They've also gone green when run locally against a bosh deployed Resolute VM.

No

@aramprice @KauzClay 

#### Summary

This PR overhauls integration-test setup/teardown to cut the number of SSH round-trips each test makes. A single spec previously issued ~148 SSH round-trips — every helper command opened a fresh SSH session through the jumpbox, so each mount, fuser, umount, losetup, etc. was a full network round-trip, and much of the cleanup ran twice per test. The suite now does far less redundant work and ships deterministic command sequences server-side, roughly halving wall-clock time on the full run.

No production agent code changes — this is entirely test-harness and CI-provisioning work.

#### What changed

Remove redundant per-test cleanup. The suite AfterEach already returns the VM to a clean baseline, but each file's BeforeEach repeated the same CleanupDataDir / CleanupLogFile / config restore. Setup/teardown now share a single RestoreCleanBaseline helper (used by both the first spec and every teardown, so the two paths can't drift), and per-file BeforeEach blocks keep only test-specific config. ResetDeviceMap becomes the single device-teardown path.

Batch deterministic command sequences into single SSH calls. CleanupDataDir, AttachDevice, ResetDeviceMap, DetachDevice, and CleanupLogFile each now fire one sudo bash -euo pipefail batch instead of N per-command round-trips, with the Go wrappers reduced to invoke-and-parse. The Noble loop-device semantics (partition auto-scan, autoclear-pending detaches, udevadm settle/retry) are preserved verbatim in the server-side script.

Hoist stable per-test setup and stream file uploads.
- Dirty-track the default agent.json so only the 3 specs that actually override it pay the restore cost (not all 49).
- Start fake-blobstore once at suite level instead of restarting it in ~15 specs; its assets now live in a dedicated dir (so SSH keys and the binary are no longer web-servable), swept clean each teardown.
- CopyFileToPath streams over the persistent SSH tunnel (sudo tee via stdin) instead of a fresh scp+mv per upload.
- Loop-device backing files are created sparsely with truncate instead of writing zeros with dd.

Fix flakes exposed by the faster suite. Tighter timing surfaced several latent races, all fixed: a root-device settle race in the ephemeral-disk spec (plus a defered restore so one failure can't cascade into a suite-wide timeout), and gate-then-separate-read TOCTOU patterns in the nats_firewall, apply, and v1_apply specs (assertions now poll a single snapshot).

Lint/format cleanup. Removed an unused type and quoted an optional scp flag (shellcheck SC2086).


docs/running_bosh_agent_integration_tests_from_local_machine.md
```diff
@@ -0,0 +1,110 @@
+# Running BOSH Agent Integration Tests From Local Machine
+
+This guide explains how to run the BOSH Agent integration tests from your local macOS machine against a remote bosh deployed VM. Be aware that running these tests shuts down the bosh agent on the VM. Turn off resurrection on the director and do not plan on `bosh ssh` working once tests are running.
+
+The integration tests (`src/bosh-agent/integration`) are primarily designed to run in CI, but you can iterate on them locally by forwarding ports and copying the necessary compiled binaries and test assets to the remote VM.
+
+## Prerequisites
+
+1.  **BOSH CLI**: You must have the BOSH CLI installed and authenticated with your environment.
```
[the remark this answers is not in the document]
mkocher: 🙄 If your director is compromised you've got way bigger problems. Keeping your bosh cli up to date is outside the scope of this development guide.

docs/running_bosh_agent_integration_tests_from_local_machine.md
```diff
@@ -0,0 +1,110 @@
...
+
+*Note: This will log you into the VM and give you a command prompt. **Leave this terminal open and running.** As long as this SSH session is active, your local port `2222` is forwarded to the VM.*
+
+### 2. Set up the `agent_test_user`
+
+The integration tests expect to run as a specific user named `agent_test_user`. Go back to your **main terminal window** and run this command to create the user and copy your SSH keys over so you can authenticate as them:
+
+```bash
+bosh -d test-bosh-agent-integration-resolute ssh -c "sudo useradd agent_test_user && sudo usermod -G bosh_sshers,bosh_sudoers agent_test_user && sudo usermod -s /bin/bash agent_test_user && sudo mkdir -p /home/agent_test_user/.ssh && sudo cp /home/vcap/.ssh/authorized_keys /home/agent_test_user/.ssh/authorized_keys && sudo chown -R agent_test_user:agent_test_user /home/agent_test_user/"
+```
```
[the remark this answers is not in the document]
mkocher: These are instructions set up to be copy and pasted once when spinning up a VM, not done multiple times.

integration/test_environment.go
```diff
@@ -101,265 +103,246 @@
...
+	const (
+		mbusUser = "mbus-user"
+		mbusPass = "mbus-pass"
+		mbusPort = 6868
+	)
+
+	lgr := logger.NewWriterLogger(logger.LevelDebug, wp)
+	mbusAddr := fmt.Sprintf("127.0.0.1:%d", mbusPort)
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
```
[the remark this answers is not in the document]
KauzClay: this is just for testing. Not in actual production

integration/test_environment.go
```diff
@@ -139,78 +138,67 @@ func NewTestEnvironment(cmdRunner boshsys.CmdRunner, wp writerPrinter) (*TestEnv
...
+// restart it: /var/log is about to become a plain directory on the root disk again, not the
+// /var/vcap/data/root_log bind mount rsyslog expects. The agent's own Bootstrap (SetupLogDir then
+// SetupLoggingAndAuditing, see agent/bootstrap.go) re-establishes that bind mount and restarts
+// logging, in that order, the next time StartAgent() runs.
+//
+// /var/log is unmounted lazily to prevent intermittent test failures: as of 2024-06-24 it is a bind
+// mount of /var/vcap/data/root_log, and for reasons we don't fully understand `fuser -k` doesn't
+// consistently terminate its holders in time for a plain umount. Because we later unmount
+// /var/vcap/data, a lingering /var/log reference will eventually fail loudly there, so this is safe.
+const detachMountFunc = `detach_mount() {
```
KauzClay: question: thoughts on making these actual sh files? Maybe in integration/assets? or a new dir?

just thinking of ease of reading/maintaining the bash

ay901246:
My only musing revolves around the shift from isolated test environments to a shared global state. Is `RestoreCleanBaseline()` enough, or is there a way to add a safety check in the teardown such that if a test leaves unexpected state behind, we don't end up chasing weird race conditions later?

KauzClay:
> My only musing revolves around the shift from isolated test environments to a shared global state. Is `RestoreCleanBaseline()` enough, or is there a way to add a safety check in the teardown such that if a test leaves unexpected state behind, we don't end up chasing weird race conditions later?

they weren't isolated test envs even before this. They all share the same bosh deployment. But also, each spec kinda had their own cleanup, which was maybe worse?

These changes just try to unify the cleanup to return the shared env to the same state after each run

docs/running_bosh_agent_integration_tests_from_local_machine.md
```diff
@@ -0,0 +1,110 @@
...
+
+Next, copy your local SSH public key to the VM so the tests can authenticate:
+
+```bash
+# Upload your public key to the VM
+bosh -d test-bosh-agent-integration-resolute scp ~/.ssh/id_rsa.pub :/tmp/id_rsa.pub
+
+# Append it to the agent_test_user's authorized_keys
+bosh -d test-bosh-agent-integration-resolute ssh -c "sudo sh -c 'cat /tmp/id_rsa.pub > /home/agent_test_user/.ssh/authorized_keys' && sudo rm /tmp/id_rsa.pub"
+```
```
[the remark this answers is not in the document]
mkocher: Fixed the comment. The intention now is to replace it.

docs/running_bosh_agent_integration_tests_from_local_machine.md
```diff
@@ -0,0 +1,110 @@
...
+
+Next, copy your local SSH public key to the VM so the tests can authenticate:
+
+```bash
+# Upload your public key to the VM
+bosh -d test-bosh-agent-integration-resolute scp ~/.ssh/id_rsa.pub :/tmp/id_rsa.pub
+
+# Append it to the agent_test_user's authorized_keys
+bosh -d test-bosh-agent-integration-resolute ssh -c "sudo sh -c 'cat /tmp/id_rsa.pub > /home/agent_test_user/.ssh/authorized_keys' && sudo rm /tmp/id_rsa.pub"
```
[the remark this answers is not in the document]
mkocher: Intention is to replace it, fixed the comment.

integration/test_environment.go
mkocher: 👍 done

ci/tasks/test-integration.sh
```diff
@@ -76,11 +76,16 @@ pushd "${bosh_agent_dir}/integration/fake-blobstore"
...
+${ssh_command} "sudo mkdir -p /home/agent_test_user/blobstore && sudo chown agent_test_user:agent_test_user /home/agent_test_user/blobstore" > /dev/null 2>&1
+
 echo -e "\n Setup assets"
 pushd "${bosh_agent_dir}/integration/assets"
-  release_folder="/home/agent_test_user/release"
+  release_folder="/home/agent_test_user/blobstore/release"
   ${ssh_command} "sudo rm -rf ${release_folder}" > /dev/null 2>&1
-  copy_to_remote_host release /home/agent_test_user/release -r
+  copy_to_remote_host release ${release_folder}" -r
 popd
```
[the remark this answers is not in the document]
mkocher: Fixed.

docs/running_bosh_agent_integration_tests_from_local_machine.md
```diff
@@ -0,0 +1,110 @@
...
+
+```bash
+bosh -d test-bosh-agent-integration-resolute ssh -c "sudo useradd agent_test_user && sudo usermod -G bosh_sshers,bosh_sudoers agent_test_user && sudo usermod -s /bin/bash agent_test_user && sudo mkdir -p /home/agent_test_user/.ssh && sudo cp /home/vcap/.ssh/authorized_keys /home/agent_test_user/.ssh/authorized_keys && sudo chown -R agent_test_user:agent_test_user /home/agent_test_user/"
+```
+
+Next, copy your local SSH public key to the VM so the tests can authenticate:
+
+```bash
+# Upload your public key to the VM
+bosh -d test-bosh-agent-integration-resolute scp ~/.ssh/id_rsa.pub :/tmp/id_rsa.pub
```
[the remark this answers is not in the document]
mkocher: Our users are smart enough to figure this out.

docs/running_bosh_agent_integration_tests_from_local_machine.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Replace the agent_test_user's authorized_keys
+bosh -d test-bosh-agent-integration-resolute ssh -c "sudo sh -c 'cat /tmp/id_rsa.pub > /home/agent_test_user/.ssh/authorized_keys' && sudo rm /tmp/id_rsa.pub"
```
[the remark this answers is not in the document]
mkocher: Fixed the description, the intention is to overwrite.

integration/scripts/cleanup_data_dir.sh
```diff
@@ -0,0 +1,50 @@
...
+sudo /var/vcap/bosh/bin/monit stop all || true
+stopped=0
+for _ in 1 2 3; do
+  summary=$(sudo /var/vcap/bosh/bin/monit summary | grep 'Process' || true)
+  total=$(printf '%s\n' "$summary" | grep -c 'Process' || true)
+  notmon=$(printf '%s\n' "$summary" | grep -c 'not monitored' || true)
+  if [ "$total" = "$notmon" ]; then stopped=1; break; fi
+  sleep 1
+done
+if [ "$stopped" != 1 ]; then echo "ensureMonitStopped: monit processes not stopped" >&2; exit 1; fi
```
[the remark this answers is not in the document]
mkocher: we can increase this if it turns out to be a problem

integration/nats_firewall_test.go
```diff
@@ -71,31 +66,30 @@ var _ = Describe("nats firewall", func() {
...
-			Expect(output).To(ContainSubstring("ct state established,related accept"))
-			Expect(output).To(MatchRegexp(`meta skuid 0 ip daddr %s tcp dport 4222 accept`, directorIP))
-			Expect(output).To(MatchRegexp(`ip daddr %s tcp dport 4222 drop`, directorIP))
+			// Assert every rule against a single snapshot, polled until all conditions hold at once, to
+			// avoid the TOCTOU race where the agent's dynamic flush+re-add of this chain drops a rule
+			// between a gating read and a separate asserting read (see the multi-url spec).
+			Eventually(func() string {
+				output, _ := testEnvironment.RunCommand("sudo nft list chain inet bosh_agent nats_access") //nolint:errcheck
+				return output
+			}, 300).Should(SatisfyAll(
```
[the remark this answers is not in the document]
mkocher: A failure will be a failure, this is a nice to have. We don't need to gold plate the failure message.

integration/compile_package_test.go
```diff
@@ -73,10 +59,7 @@ var _ = Describe("compile_package", func() {
 		multiDigest := createSHA1MultiDigest("236cbd31a483c3594061b00a84a80c1c182b3b20")
 
 		BeforeEach(func() {
-			err := testEnvironment.StartBlobstore()
-			Expect(err).NotTo(HaveOccurred())
-
-			err = testEnvironment.CopyFileToPath(filepath.Join(testEnvironment.AssetsDir(), "dummy_package.tgz"), filepath.Join(testEnvironment.BlobstoreDir(), "dummy_package.tgz"))
+			err := testEnvironment.CopyFileToPath(filepath.Join(testEnvironment.AssetsDir(), "dummy_package.tgz"), filepath.Join(testEnvironment.BlobstoreDir(), "dummy_package.tgz"))
 			Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
mkocher: Clod Says: the premise "the agent cleans up dirs on start" doesn't apply to this destination. The copy target is `/home/agent_test_user/blobstore`. The fixture staged in step 2 survives StartAgent in step 3 and is served to the agent via the signed URL during the It body. The test is correct as written. Moving the copy to a nested JustBeforeEach would be a harmless no-op change based on a misread of where dummy_package.tgz lands — I'd decline it.

infrastructure/settings_source_factory.go
```diff
@@ -166,6 +166,52 @@ func (f SettingsSourceFactory) buildWithoutRegistry() (boshsettings.Source, erro
 	return NewMultiSettingsSource(f.logger, settingsSources...)
 }
 
+// MarshalJSON is the inverse of UnmarshalJSON: it re-injects the "Type" discriminator that the
+// concrete SourceOptions structs (FileSourceOptions, HTTPSourceOptions, ...) don't carry as a field,
+// so a marshaled slice round-trips back through UnmarshalJSON. Without it, json.Marshal emits source
+// objects with no "Type" and reloading fails with "Missing source type".
+func (s SourceOptionsSlice) MarshalJSON() ([]byte, error) {
```
Alphasite, line 173: nice! very rare to see someone making good use of this feature

infrastructure/settings_source_factory_test.go
```diff
@@ -235,4 +235,51 @@ var _ = Describe("SettingsSourceFactory", func() {
...
+				"ConfigDrive":      ConfigDriveSourceOptions{DiskPaths: []string{"/dev/vdb"}},
+				"File":             FileSourceOptions{SettingsPath: "/tmp/settings.json"},
+				"CDROM":            CDROMSourceOptions{FileName: "env"},
+				"VsphereGuestInfo": VsphereGuestInfoSourceOptions{},
+			}
+
+			for wantType, opts := range cases {
+				data, err := json.Marshal(SourceOptionsSlice{opts})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(string(data)).To(ContainSubstring(`"Type":"` + wantType + `"`))
```
Alphasite, line 270: this is fine, but iirc gomega has a nice json matcher for this exact use case which is even better

integration/agent_config.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/cloudfoundry/bosh-agent/v2/app"
+	boshinf "github.com/cloudfoundry/bosh-agent/v2/infrastructure"
+	boshplatform "github.com/cloudfoundry/bosh-agent/v2/platform"
+)
+
+// ServiceManager is intentionally left zero here; CreateAgentConfigFile stamps it per target
+// (sv vs systemd) so this value stays service-manager-agnostic.
+var DefaultAgentConfig = app.Config{
```
Alphasite, line 11: i cant remember the rules here but can you make this const?

integration/apply_test.go
```diff
@@ -78,15 +75,10 @@ var _ = Describe("apply", func() {
...
-	AfterEach(func() {
-		err := testEnvironment.DetachDevice("/dev/sdh")
-		Expect(err).ToNot(HaveOccurred())
-	})
-
 	JustBeforeEach(func() {
 
-		// StartAgentTunnel also acts as a wait condition for the agent to have fully started. Copying over the blobs before it fully starts, will result in issues because the agent cleans up dirs on start.
-		err := testEnvironment.StartAgentTunnel()
+		// WaitForAgent blocks until the agent is reachable. Copying over the blobs before it fully starts will result in issues because the agent cleans up dirs on start.
```
Alphasite, line 80: nit: line length

integration/apply_test.go
```diff
@@ -154,21 +146,20 @@ var _ = Describe("apply", func() {
...
+		for _, dir := range []string{
+			"/var/vcap/data/sys/run/foobar",
+			"/var/vcap/data/sys/log/foobar",
+			"/var/vcap/data/foobar",
+		} {
+			dir := dir
+			Eventually(func() string {
+				out, _ := testEnvironment.RunCommand("sudo stat " + dir) //nolint:errcheck
+				return out
+			}, 2*time.Minute, 1*time.Second).Should(MatchRegexp("Access: \\(0770/drwxrwx---\\)  Uid: \\(    0/    root\\)   Gid: \\( 100[0-9]/    vcap\\)"))
```
Alphasite, line 162: The looooong regex just makes the assertion hard to read (heres my possibly bad suggestion)

```suggestion
			Eventually(
			    func() string {
					out, _ := testEnvironment.RunCommand("sudo stat " + dir) //nolint:errcheck
					return out
   			},
   			2*time.Minute, 
   			1*time.Second,
			).Should(
				MatchRegexp("Access: \\(0770/drwxrwx---\\)  Uid: \\(    0/    root\\)   Gid: \\( 100[0-9]/    vcap\\)"),
			)
```
or something like that (not sure what the patch its going to suggest looks like exactly.

integration/bundle_logs_test.go
```diff
@@ -50,16 +47,13 @@ var _ = Describe("bundle_logs", func() {
...
-		err := testEnvironment.StartAgentTunnel()
+		err := testEnvironment.WaitForAgent()
 		Expect(err).NotTo(HaveOccurred())
 	})
 
 	AfterEach(func() {
 		_, err := testEnvironment.RunCommand("sudo userdel -rf username || true") //nolint:errcheck
 		Expect(err).ToNot(HaveOccurred())
-
-		err = testEnvironment.DetachDevice("/dev/sdh")
```
Alphasite, line 61: im missing some context but who calls the detach scripts now?

integration/compile_package_test.go
```diff
@@ -23,15 +23,6 @@ var _ = Describe("compile_package", func() {
 	)
 
 	BeforeEach(func() {
-		err := testEnvironment.CleanupDataDir()
```
Alphasite, line 26: q: as above who calls these scripts now? WaitForAgent or something?

integration/nats_firewall_test.go
```diff
@@ -71,31 +66,30 @@ var _ = Describe("nats firewall", func() {
...
-				output, err := testEnvironment.RunCommand("sudo nft list chain inet bosh_agent nats_access")
-				return err == nil && strings.Contains(output, directorIP)
-			}, 300).Should(BeTrue())
-
-			output, err := testEnvironment.RunCommand("sudo nft list chain inet bosh_agent nats_access")
-			Expect(err).To(BeNil())
-			Expect(output).To(ContainSubstring("ct state established,related accept"))
-			Expect(output).To(MatchRegexp(`meta skuid 0 ip daddr %s tcp dport 4222 accept`, directorIP))
-			Expect(output).To(MatchRegexp(`ip daddr %s tcp dport 4222 drop`, directorIP))
+			// Assert every rule against a single snapshot, polled until all conditions hold at once, to
```
Alphasite, line 76: nice

Alphasite (commented):
didnt have time to finish the review but left some comments

integration/bundle_logs_test.go
mkocher: they happen in a suite level AfterEach that cleans up everything

integration/compile_package_test.go
mkocher: there's an AfterEach in integration_suite_test.go which always calls RestoreCleanBaseline() which cleans up everything

KauzClay: the cleanup got folded into a function called `RestoreCleanBaseline()` that gets called in the suite-level AfterEach ([hopefully this link works](https://github.com/cloudfoundry/bosh-agent/pull/457/changes#diff-eef4eac5433048ae80185dfd942bbd2303d32847592396ff663eb8d4c25c9e58R70))

integration/bundle_logs_test.go
KauzClay: the cleanup got folded into a function called RestoreCleanBaseline() that gets called in the suite-level AfterEach ([hopefully this link works](https://github.com/cloudfoundry/bosh-agent/pull/457/changes#diff-eef4eac5433048ae80185dfd942bbd2303d32847592396ff663eb8d4c25c9e58R70))

integration/agent_config.go
KauzClay: no I don't think you can make structs consts

infrastructure/settings_source_factory_test.go
KauzClay: cool, I didn't know about MatchJSON, if that is what you were referring to.

Based on reading how MatchJSON works [here](https://pkg.go.dev/github.com/onsi/gomega#MatchJSON), I'm inclined to say we just keep this explicit check for the type field though.

I also think MatchJSON doesn't do partial matching, so we'd need to write out a full string each time, something like: `{"Headers":null,"InstanceIDPath":"","SSHKeysPath":"","TokenPath":"","Type":"HTTP","URI":"http://example.com","UserDataPath":""}`

KauzClay (approved):

PR #458, opened 2026-08-03
refactor: define agent config in code

KauzClay:

after adding copies of the agent config that used systemd, it seemed execessive to have 6 json config files used by tests

instead, we can just define a default config in code, let the test harnesss determine if sysetmd is being used, and edit the agent config as needed per test

ai-assisted=yes

[TNZ-106038] Refactor bosh-agent CI pipeline to run integration tests on all supported stemcell lines (Jammy, Noble, Resolute)

TODO

N/A

No

@mkocher 

### AI Review Feedback

_All AI review comments (CodeRabbit, and Copilot when assigned) must be resolved before human reviewers will look at the PR. For each comment, either:_
- _Make the suggested change, or_
- _Reply to the comment explaining why it does not apply, then resolve the thread._


integration/agent_config.go
```diff
@@ -0,0 +1,25 @@
+package integration
+
+import (
+	"github.com/cloudfoundry/bosh-agent/v2/app"
+	boshinf "github.com/cloudfoundry/bosh-agent/v2/infrastructure"
+	boshplatform "github.com/cloudfoundry/bosh-agent/v2/platform"
+)
+
+var DefaultAgentConfig = app.Config{
```
KauzClay, line 9: question: should this be defined in test_environment.go ?

That seems to be where most of the test helper code is

integration/ephemeral_disk_test.go
```diff
@@ -83,7 +83,9 @@ var _ = Describe("EphemeralDisk", func() {
 
 				Context("when bind mount /var/vcap/data/root_tmp on /tmp", func() {
 					BeforeEach(func() {
-						err := testEnvironment.UpdateAgentConfig(testEnvironment.GetSettingsFile("no-default-tmp-dir"))
+						agentConfig := integration.DefaultAgentConfig
+						agentConfig.Platform.Linux.UseDefaultTmpDir = false
+						err := testEnvironment.CreateAgentConfigFile(agentConfig)
```
KauzClay: nit: could the `CreateAgentConfigFile(agentConfig)` be put in a top-level just before each or something, along with `agentConfig := integration.DefaultAgentConfig`

and then the contexts could modify the value for `agentConfig` as needed?

integration/test_environment.go
```diff
@@ -691,12 +692,24 @@ func (t *TestEnvironment) TearDownDummyNetworkInterface() error {
 	return err
 }
 
-func (t *TestEnvironment) UpdateAgentConfig(configFile string) error {
-	_, err := t.RunCommand("sudo rm -f /var/vcap/bosh/agent.json")
+func (t *TestEnvironment) CreateAgentConfigFile(config app.Config) error {
```
KauzClay, line 695: nit: this can probably still be called UpdateAgentConfig, no?

integration/test_environment.go
```diff
@@ -691,12 +692,24 @@ func (t *TestEnvironment) TearDownDummyNetworkInterface() error {
...
+	if config.Platform.Linux.ServiceManager == "" {
+		config.Platform.Linux.ServiceManager = t.serviceManager
+	}
```
KauzClay: nit: do we even need this if statement? I think t.serviceManager will return "" anyway if the stemcell isn't using systemd.

could it just be `config.Platform.Linux.ServiceManager = t.serviceManager`?

KauzClay:
probably should rebase this on top of https://github.com/cloudfoundry/bosh-agent/pull/457

KauzClay:
closing because we are folding this change into https://github.com/cloudfoundry/bosh-agent/pull/457

PR #464, opened 2026-08-06
Re-attempt GrowFilesystem when partition already matches disk size

neddp:

When `AdjustPersistentDiskPartitioning` is called after a disk resize, it either resizes the partition and grows the filesystem (if the partition is smaller than the disk), or falls into an `else` branch that only partitions and formats (if the partition already spans the disk). The `else` branch never called `GrowFilesystem`.

This means: if a previous deploy resized the partition but `GrowFilesystem` failed, every subsequent deploy would see the partition already at full size, skip the grow silently, and succeed - leaving the filesystem permanently smaller than its partition with no visible error.

Fix: call `GrowFilesystem` unconditionally in the `else` branch after `Format`. `resize2fs` and `xfs_growfs` are idempotent and exit 0 when the filesystem already fills the partition, so this is safe on every deploy. If the grow fails, the error is returned and the deployment fails, forcing the operator to resolve it.

Discovered during an investigation into a `resize2fs: Permission denied` failure. Three nodes had pre-existing filesystem corruption from a storage write-error event. The partition resize succeeded but `GrowFilesystem` failed. Subsequent deploys silently succeeded, leaving those nodes with 98G filesystems on 1T volumes.

- Full `platform` unit test suite: 373 passed, 0 failed
- New specs covering: grow called in else branch, grow failure returns error

When a persistent disk's partition already matches the disk size but the filesystem is smaller (due to a previously failed grow), the agent now re-attempts the filesystem grow on each deploy instead of silently skipping it.

No. `resize2fs`/`xfs_growfs` are idempotent on already-full filesystems. The only behaviour change is that a previously silent failure now surfaces as a deploy error.

### DIFF
--- a/platform/linux_platform.go
+++ b/platform/linux_platform.go
@@ -1451,6 +1451,11 @@ func (p linux) AdjustPersistentDiskPartitioning(diskSetting boshsettings.DiskSet
 		if err != nil {
 			return bosherr.WrapError(err, fmt.Sprintf("Formatting partition with %s", diskSetting.FileSystemType))
 		}
+
+		// Grow is idempotent — re-attempt on every run to catch incomplete prior grows.
+		if err = p.diskManager.GetFormatter().GrowFilesystem(firstPartitionPath); err != nil {
+			return bosherr.WrapError(err, "Failed to grow filesystem")
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


neddp:
Not a good implementation.

PR #465, opened 2026-08-06
Warn when persistent disk filesystem is smaller than its partition

neddp:

When a persistent disk is resized, `AdjustPersistentDiskPartitioning` resizes the partition and grows the filesystem. If the partition resize succeeds but the filesystem grow does not complete, every subsequent deploy sees the partition already spanning the disk, takes the branch that only partitions and formats, and never revisits the filesystem - leaving it silently smaller than its partition with no indication anything is wrong.

This change adds a `FilesystemNeedsGrow` check (ext4) to the formatter. When the partition already matches the disk size, `AdjustPersistentDiskPartitioning` runs the check and logs a warning if the filesystem is smaller than its partition, surfacing the condition for operator follow-up.

Detection only - no automatic grow is attempted:
- The filesystem is not mounted at this point (`AdjustPersistentDiskPartitioning` runs before `MountPersistentDisk`), so `xfs_growfs`, which requires a mounted filesystem, is not applicable.
- The underlying cause often requires manual intervention (e.g. a filesystem flagged with errors that the kernel refuses to grow online).

`FilesystemNeedsGrow` compares the ext4 superblock size (`dumpe2fs -h`) against the block device size (`blockdev --getsize64`), using the same 100MB delta as `SinglePartitionNeedsResize` to avoid false positives from alignment rounding.

Found while investigating a `resize2fs: Permission denied` failure. Three nodes had pre-existing filesystem corruption. The partition resize succeeded, `GrowFilesystem` failed, and subsequent deploys silently succeeded - leaving those nodes with 98G filesystems on 1T volumes and no visible error.

- Full `platform` and `platform/disk` unit test suites: 375 + 177 passed, 0 failed
- New specs: `FilesystemNeedsGrow` for ext4 (smaller, already-full, command failures, missing superblock field, non-ext4); platform specs for the warn path (check invoked, no grow/mount, no deploy failure on mismatch or check error)

The agent now logs a warning when a persistent disk's filesystem is smaller than its partition (indicating a previous grow did not complete), instead of silently leaving it unresized.

No. The change only adds a read-only check and a log warning. No behaviour change to partitioning, formatting, or mounting.

### DIFF
--- a/platform/disk/fakes/fake_formatter.go
+++ b/platform/disk/fakes/fake_formatter.go
@@ -14,6 +14,11 @@ type FakeFormatter struct {
 	GrowFilesystemCalled        bool
 	GrowFilesystemPartitionPath string
 	GrowFilesystemError         error
+
+	FilesystemNeedsGrowCalled        bool
+	FilesystemNeedsGrowPartitionPath string
+	FilesystemNeedsGrowResult        bool
+	FilesystemNeedsGrowError         error
 }
 
 func NewFakeFormatter() *FakeFormatter {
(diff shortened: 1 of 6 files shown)


neddp:
The implementation is too complicated. Not worth the technical debt.

PR #470, opened 2026-08-19
Group static interface configurations per interface

wayneeseguin:
## Problem

Two manual networks bound to one NIC via `nic_groups` each produce a `StaticInterfaceConfiguration` with the same interface name. `writeNetworkInterfaces` renders each one to the same systemd-networkd unit, `10_<iface>.network`, so the last write wins and the interface comes up with only one of its static addresses. Which address survives depends on map iteration order. The agent's own static address validation (`InterfaceAddressesValidator`) then fails on the address that lost, and bootstrap never completes — the director times out pinging the agent.

Commit c83e3f85 (RFC 0038) grouped the **DHCP** configurations per interface, but the **static** configurations stayed at one file per configuration.

## Reproduction

Any deployment that binds a manual IPv4 network and a manual IPv6 network to one NIC with `nic_groups`, e.g. the BATS example `supports manual network dual stack` (`ipv6_spec.rb`). Observed on a Proxmox VE CPI BATS run with bosh-stemcell `ubuntu-noble` 1.383 and director 282.1.13:

- cloud-init put both families on the NIC (`ip=10.254.48.190/16,gw=10.254.0.1,ip6=fd36:afd4:2c42:1::48:190/64,gw6=fd36:afd4:2c42:1::1`)
- `settings.json` carried both networks with the same MAC
- the agent wrote a single `/etc/systemd/network/10_eth0.network` containing only the IPv4 address, failed `Validating static network configuration`, and the deploy timed out after 600s

The same collision hits two manual IPv4 networks sharing a NIC (`multiple_manual_networks` with a shared `nic_group`).

## Fix

Group static configurations by interface name, exactly as c83e3f85 did for DHCP, and render one unit per interface:

- one `[Address]` section per configuration — IPv4 before IPv6, ordered by address within each family, so the rendered file is stable across agent runs (and `ConvergeFileContents` doesn't see spurious changes)
- `Gateway=` for each configuration marked default for the gateway; `IPv6AcceptRA=true` when any configuration on the interface is IPv6
- `[Route]` sections merged from every configuration
- virtual interface (alias) `[Address]` sections deduplicated, since `collapseVirtualInterfaces` attaches the same list to every configuration sharing the device

A single configuration per interface renders byte for byte the same unit as before — all existing specs pass unchanged.

## Testing

- new spec: manual IPv4 + manual IPv6 sharing a MAC renders one unit with both `[Address]` sections (fails on main: the unit holds only one address, nondeterministically)
- new spec: two manual IPv4 networks sharing a MAC — deterministic address order, merged `[Route]` sections, gateway from the default-for-gateway network only
- `bin/ginkgo --skip-package=integration` green; `golangci-lint run ./platform/net/...` clean for `GOOS=linux` and `GOOS=windows`; both cross-builds compile


wayneeseguin:
I ran into this because with our proxmoxm cpi effort we are trying to get BATS running with dual stack for it.

neddp (approved):
I believe this might also be addressing the issue with the Azure implementation. Did not confirm.
https://github.com/cloudfoundry/bosh-azure-cpi-release/pull/746

Thanks!

PR #471, opened 2026-08-20
Fix Lint issues after new Go version

KauzClay:

Addressing the errors that showed up in https://ci.bosh-ecosystem.cf-app.com/teams/main/pipelines/bosh-agent-main/jobs/test-unit/builds/889

We recently bumped our go version, and coincidentally, the golangci-lint tool, which flagged some new things.

Ran locally
```
❯ ./bin/lint
golangci-lint has version 2.13.1 built with go1.26.7 from (unknown, modified: ?, mod sum: "h1:RuM4OcluM4xFQcGuRE6R7jA33pqxK/W1EsBxpugdZjg=") on (unknown)

 lint-ing with GOOS=windows...
0 issues.

 lint-ing with GOOS=linux...
0 issues.

 shellcheck not found; install it (https://github.com/koalaman/shellcheck#installing)
```

N/A

No

N/A

### AI Review Feedback

_All AI review comments (CodeRabbit, and Copilot when assigned) must be resolved before human reviewers will look at the PR. For each comment, either:_
- _Make the suggested change, or_
- _Reply to the comment explaining why it does not apply, then resolve the thread._

### DIFF
--- a/integration/utils/blobstore.go
+++ b/integration/utils/blobstore.go
@@ -65,7 +65,7 @@ func NewBlobstore(uri string) BlobClient {
 
 	httpClient := &http.Client{
 		Transport: &http.Transport{
-			Dial:                tunnelClient.Dial,
+			DialContext:         tunnelClient.DialContext,
 			TLSHandshakeTimeout: 10 * time.Second,
 		},
 	}
--- a/platform/firewall/nftables_firewall_other.go
+++ b/platform/firewall/nftables_firewall_other.go
@@ -8,7 +8,9 @@ import (
 	boshlog "github.com/cloudfoundry/bosh-utils/logger"
 )
 
+var errNftablesUnsupported = errors.New("nftables firewall is only supported on Linux")
+
 // NewNftablesFirewall returns an error on non-Linux platforms
 func NewNftablesFirewall(logger boshlog.Logger) (Manager, error) {
-	return nil, errors.New("nftables firewall is only supported on Linux")
+	return nil, errNftablesUnsupported
 }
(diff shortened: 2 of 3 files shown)


integration/utils/blobstore.go
```diff
@@ -65,7 +67,9 @@ func NewBlobstore(uri string) BlobClient {
 
 	httpClient := &http.Client{
 		Transport: &http.Transport{
-			Dial:                tunnelClient.Dial,
+			DialContext: func(ctx context.Context, network string, addr string) (net.Conn, error) {
```
ay901246: nit - Feels a little awkward to add all of this just to drop the ctx, I guess it could be replaced with `_ context.Context` instead as a small improvement?

aramprice (dismissed):

platform/firewall/nftables_firewall_other.go
```diff
@@ -8,7 +8,9 @@ import (
 	boshlog "github.com/cloudfoundry/bosh-utils/logger"
 )
 
+var errNftablesUnsupported = errors.New("nftables firewall is only supported on Linux")
```
Alphasite, line 11: question: should this just be a `fmt.errorf(...)` instead? i thought this method was only for type switching/assertions (effectively)

Alphasite (dismissed):
Fine, just q about the errors.new vs fmt.errorf

ay901246 (dismissed):

platform/firewall/nftables_firewall_other.go
aramprice: I think `fmt.Errorf()` is no good if the string isn't being interpolated.

integration/utils/blobstore.go
aramprice: Would it be cleaner (possible?) to do:

```golang
	httpClient := &http.Client{
		Transport: &http.Transport{
			DialContext:         tunnelClient.DialContext,
			TLSHandshakeTimeout: 10 * time.Second,
		},
	}
```

KauzClay: Hmmm good points. I'll try it out tomorrow

KauzClay:
merging on coderabbit's approval since I got the human approvals earlier, and the new change that wiped away approvals just addresses a nit.

PR #473, opened 2026-08-28
Release the SCSI device when a persistent disk is detached

julian-hj:
This resolves a new issue surfaced in vSphere 9.1 where attaching orpahned disks or repeatedly resizing a disk fails.

remove_persistent_disk only dropped the disk's settings hints, leaving the kernel's device object in place. After the IaaS-side detach that object is dead — I/O returns DID_NO_CONNECT — but it still occupies the disk's controller slot. When the IaaS later reuses that slot for another disk, the kernel does not re-probe an address it believes is populated, so the incoming disk is never enumerated, no /dev/disk/by-id entry appears, and mounting fails with "Timed out getting real device path".

A SCSI rescan does not help: SCSIIDDevicePathResolver already performs one, and it only adds devices at unoccupied addresses.

This breaks routine deploys, not just recovery workflows. Resizing a persistent disk detaches the old disk from a running VM, so every resize strands a dead device. update_disk attaches the replacement before detaching the old disk, so the first resize survives — the old disk still holds its slot and the new one lands elsewhere. The second resize on the same VM is then handed the slot the first one stranded, and fails. `bosh attach-disk` onto an instance that already has a persistent disk fails the same way, on its first use, because AttachDisk detaches before attaching.

The director sends remove_persistent_disk from DetachDiskStep immediately before calling the CPI, so the action now releases the device there, while the disk is still attached. That is the documented order for removing a SCSI disk from a Linux guest: unmount, let the kernel release the device, then detach at the IaaS. DetachDiskStep backs every detach path, so this covers resize and attach-disk alike.

Device removal is best effort and idempotent. A disk with no recorded settings has no device to release, which is normal when a detach is retried or rolled back, so that is not an error. The device is resolved with a single non-blocking lookup rather than through devicePathResolver, whose SCSI timeout is 50s, because this runs on paths where the device is legitimately already gone.

Reproduced and verified on vCenter 9.1, where vm-operator assigns the lowest free controller unit and so reuses a stranded slot. Before: 1GiB -> 2GiB resize left a dead device, and the 2GiB -> 3GiB resize failed; attach-disk failed every attempt. After: the resize chain completes, the replacement disk mounts on the reused slot, and the guest is left with no dead devices.

### DIFF
--- a/agent/action/remove_persistent_disk.go
+++ b/agent/action/remove_persistent_disk.go
@@ -5,16 +5,19 @@ import (
 
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 
+	boshplatform "github.com/cloudfoundry/bosh-agent/v2/platform"
 	boshsettings "github.com/cloudfoundry/bosh-agent/v2/settings"
 )
 
 type RemovePersistentDiskAction struct {
 	settingsService boshsettings.Service
+	platform        boshplatform.Platform
 }
 
-func NewRemovePersistentDiskAction(settingsService boshsettings.Service) RemovePersistentDiskAction {
+func NewRemovePersistentDiskAction(settingsService boshsettings.Service, platform boshplatform.Platform) RemovePersistentDiskAction {
 	return RemovePersistentDiskAction{
 		settingsService: settingsService,
+		platform:        platform,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


agent/action/remove_persistent_disk.go
```diff
@@ -24,6 +27,24 @@ func (a RemovePersistentDiskAction) Run(diskCID string) (interface{}, error) {
 		return nil, bosherr.WrapError(err, "Refreshing the settings")
 	}
 
+	// The director sends this immediately before asking the CPI to detach the
```
selzoc, line 30: Do any other workflows that aren't detaching the disk use this path?

selzoc (approved):

KauzClay (approved):

Issue #428, opened 2026-05-07
Intermittent bootstrap failure on AWS Nitro instances


Ivaylogi98:
## Summary

On AWS Nitro-based instances with an encrypted ephemeral EBS disk, the bosh-agent intermittently fails to bootstrap with `timed out pinging VM` or `timed out sending get_task to instance`.

The agent attempts to partition the root disk instead of the ephemeral disk, which causes the bootstrap to fail.

## Environment

- bosh v282.1.2
- bosh-aws-cpi v107.0.1
- Instance type: `t3a.nano` (Nitro, no local instance storage)
- Cloud config:
  ```yaml
  cloud_properties:
    ephemeral_disk:
      encrypted: true
      size: 20000
    instance_type: t3a.nano

## Observed behaviour

- Intermittent — not consistently reproducible, timing-dependent
- Agent attempts to partition the root disk
- VM becomes unresponsive, deployment times out

## Analysis

On Nitro instances, all disks appear as NVMe devices. The agent resolves /dev/sdb (the ephemeral disk name in settings) to an NVMe device via /dev/disk/by-id/ symlinks using IDDevicePathResolver. It is unclear whether encrypted EBS disks produce a different symlink name in /dev/disk/by-id/ than what the agent's glob pattern expects, or whether the disk ID passed from the CPI does not match the symlink.

## Related

- cloudfoundry/bosh-agent#407 — fixes the same class of problem for NVMe instance
  storage discovery
- cloudfoundry/bosh-aws-cpi-release#196 — companion CPI-side change, may also be relevant here

Co-authored with Claude Code

paul-szymanski-a9s:
_Disclaimer: This was fully investigated by AI and I did not have the capacity to review this:_

We hit this on an independent environment and traced a concrete mechanism. Summary up front: the symptom reproduces exactly, but I believe the analysis in the issue description is looking at the wrong resolver — in this configuration `IDDevicePathResolver` and `/dev/disk/by-id/` are never involved in ephemeral disk resolution. The suspect is `MappedDevicePathResolver`'s 30 s timeout racing AWS NVMe udev symlink creation, followed by a **silent fallback that partitions the root disk**.

## Environment

- BOSH director 282.1.13, `aws_cpi`
- Stemcell `bosh-aws-xen-hvm-ubuntu-jammy-go_agent/1.1298` (ubuntu-jammy)
- Instance type `t3a.small` (Nitro, no local instance storage)
- ```yaml
[...]

PR #480, opened 2026-09-16
Fix silent ephemeral disk fallback and make resolver timeout configurable

Ivaylogi98:

On AWS Nitro instances with an encrypted ephemeral EBS disk, the agent could silently fall back to partitioning the root disk when the mapped device path resolver timed out before the `/dev/sd*` symlink appeared. This happened because the resolver had a hardcoded 30 s timeout, and on encrypted EBS volumes the device can take longer to become visible.

This PR makes two related fixes:

1. **Configurable timeout** — `LinuxOptions.EphemeralDiskWaitTimeout` (seconds) lets operators tune the wait via the agent config. Falls back to a new default of 60 s (double the previous 30 s hardcoded value). Applies to the `virtio` resolver path on AWS.
2. **Fail loudly on resolution failure** — when a `DiskSettings.Path` was provided by the CPI but the resolver fails, `GetEphemeralDiskPath` now returns an error instead of silently returning an empty string. The silent empty-string return was the root cause of the agent falling through to root-disk partitioning. When no path is provided (no ephemeral disk configured), the existing behaviour (return `""`, no error) is preserved so the root partition path still works.

- Issue: https://github.com/cloudfoundry/bosh-agent/issues/428
- Related: cloudfoundry/bosh-agent#407 (same class of problem for NVMe instance storage discovery)
- Related: cloudfoundry/bosh-aws-cpi-release#196 (companion CPI-side change for #407)

- Unit tests: `go test ./platform/...` — all passing
- New `platform/provider_internal_test.go` covers `mappedDevicePathResolverTimeout`: configured value, unset fallback, negative value fallback, and assertion that the default is larger than the historical 30 s
- Updated `linux_platform_test.go` covers both the error-is-propagated (path provided) and error-is-suppressed (no path) branches of `GetEphemeralDiskPath`

Operators can now configure how long the agent waits for a mapped ephemeral disk device (e.g. `/dev/sdb`) to appear, via `EphemeralDiskWaitTimeout` (seconds) in `LinuxOptions`. The default has been raised from 30 s to 60 s. The agent now also fails loudly — instead of silently falling back to the root disk — when a CPI-provided ephemeral disk path cannot be resolved within the timeout.

No for the configurable timeout — the default is more permissive than before.

The fail-loudly change is technically a behaviour change: deployments that previously silently fell back to root-disk partitioning will now fail with an explicit error. This is the intended outcome; the silent fallback was masking a real misconfiguration or timing issue.

### AI Review Feedback

_All AI review comments (CodeRabbit, and Copilot when assigned) must be resolved before human reviewers will look at the PR._


platform/linux_platform.go
```diff
@@ -103,6 +103,10 @@ type LinuxOptions struct {
 	// Base path for LUN-based symlink resolution (e.g., "/dev/disk/azure/data/by-lun").
 	LunDeviceSymlinkPath string
 
+	// Number of seconds the mapped device path resolver waits for a /dev/sd* ephemeral/persistent
+	// device (or its compatibility symlink) to appear before giving up.
+	EphemeralDiskWaitTimeout int
```
yuriadam-sap: Maybe just a nit pick, but `EphemeralDiskWaitTimeout` sounds like it applies to all ephemeral disk resolution, but it only affects the mapped device path resolver inside the `virtio` path. 

Would something like `VirtioEphemeralDiskWaitTimeout` be clearer, or maybe just a short comment on the field noting it's virtio-only?

Ivaylogi98: Good suggestion, renamed it to VirtioDevicePathResolverTimeout as the mapped resolver resolves both ephemeral and persistent. Fixed with [90c1613](https://github.com/cloudfoundry/bosh-agent/pull/480/commits/90c161390d4c8a81fd58b0b86654669bb49d486c)

yuriadam-sap (approved):

